PR #14, opened 2019-06-20
Update README.md

manhhung741:
Add codecov badge


manhhung741:
@nacin The codecov report looks fishy.

PR #23, opened 2019-06-22
Try to more accurately map column types to ActiveRecord Types

pje:
Hey @manhhung741! 👋 Thanks for all your work on this project, and for open-sourcing it!

This changeset tries to be more accurate in generating Sorbet types for model getters and setters by mapping the `ActiveRecord::Type` class to specific Sorbet types.

Currently, for example, `decimal` columns always generate method signatures that operate on `Integer`s, but you'd almost always want them to be `BigDecimal`s.

I've tried to mirror Rails's type casting behavior as closely as possible here, but it's possible I've missed some finer details.


spec/model_rbi_formatter_spec.rb
```diff
@@ -37,8 +37,7 @@ def expect_match_file(content, file_path, save_to_file: false)
     formatter = ModelRbiFormatter.new(Wand, ['Wizard', 'Wand', 'SpellBook'])
     expect_match_file(
       formatter.generate_rbi,
-      'spec/test_data/models/expected_wand.rbi',
-      save_to_file: true,
+      'spec/test_data/models/expected_wand.rbi'
```
pje: It seemed strange to me that we were always overwriting the expectation file during tests—I couldn't get the test to fail at first without removing `save_to_file` here!

manhhung741: Oh thanks, this is a bug. The `save_to_file` flag is to regenerate expected output file.

spec/support/rails_app/db/schema.rb
```diff
@@ -22,6 +22,9 @@
     t.integer "wizard_id", null: false
     t.string "wood_type"
     t.integer "core_type"
+    t.float "flexibility", null: false
```
manhhung741: XD these would make my teammates really happy. 
cc @liza8bit @fresh-prince-of-prosemirror

spec/model_rbi_formatter_spec.rb
manhhung741: Can you add a comma at the end of the line?

lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -285,13 +272,40 @@ def generate_column_methods(buffer)
...
     if column_def.respond_to?(:array?) && column_def.array?
       strict_type = "T::Array[#{strict_type}]"
     end
     column_def.null ? "T.nilable(#{strict_type})" : strict_type
   end
 
+  def active_record_type_to_sorbet_type(klass)
+    case klass
+    when ActiveRecord::Type::Json
+      T.any(Array, Float, Hash, Integer, NilClass, String)
```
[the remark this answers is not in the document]
manhhung741: I'd also add a non-nulable `t.json "maker_info"` to test the case that a json is not null

spec/support/rails_app/db/schema.rb
```diff
@@ -22,6 +22,9 @@
...
+    t.decimal "hardness", null: false, precision: 10, scale: 10
+    t.decimal "reflectance", null: false, precision: 10, scale: 0
```
manhhung741:
```suggestion
    t.decimal "reflectance", null: false, precision: 10, scale: 0
    t.date "chosen_at_date"
    t.time "chosen_at_time"
```

manhhung741:
I convert some requested changes to the form of suggestions

spec/support/rails_app/db/schema.rb
pje: Will do! Also I forgot to add `T::Boolean` as a possible JSON value type. Will add that too.

pje: Alright, just added the migration and support for `T::Boolean` in json columns.

PR #26, opened 2019-06-25
rename rbis/ folder to rbi/ because sorbet drop support for rbis/

manhhung741:
In the latest version of sorbet, "srb tc" works automatically.


nacin:
See sorbet/sorbet#1037.

PR #27, opened 2019-06-25
Don't raise an error when processing an abstract_class. Also: blacklist ApplicationRecord in rake task

pje:
Currently, when generating model rbis from the rake task, we try to generate rbis for `ApplicationRecord`—which is an abstract_class that we should expect to not have a table—and we print a confusing error message:

```
Error when handling model ApplicationRecord: Mysql2::Error: Incorrect table name '': SHOW FULL FIELDS FROM ``
```

Seems like we should just filter out `ApplicationRecord` in the rake task. What do you think, @manhhung741?


lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -16,14 +16,14 @@ class ModelRbiFormatter
...
+    @columns_hash = model_class.table_exists? ? model_class.columns_hash : {}
     @generated_sigs = ActiveSupport::HashWithIndifferentAccess.new
     @generated_class_sigs = ActiveSupport::HashWithIndifferentAccess.new
     @generated_scope_sigs = ActiveSupport::HashWithIndifferentAccess.new
     @generated_querying_sigs = ActiveSupport::HashWithIndifferentAccess.new
     begin
       # Load all dynamic instance methods of this model by instantiating a fake model
-      @model_class.new
+      @model_class.new unless @model_class.abstract_class?
     rescue StandardError
```
pje: Also: TIL that `NotImplementedError` does not inherit from `StandardError`.

lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -21,6 +21,7 @@ namespace :rails_rbi do
 
     blacklisted_models = []
     blacklisted_models << ActiveRecord::SchemaMigration if defined?(ActiveRecord::SchemaMigration)
+    blacklisted_models << ApplicationRecord if defined?(ApplicationRecord)
```
manhhung741: I honestly thought I blacklisted ApplicationRecord.

spec/model_rbi_formatter_spec.rb
```diff
@@ -25,6 +25,11 @@ def expect_match_file(content, file_path, save_to_file: false)
     ])
   end
 
+  it 'does not throw an error when given an abstract class' do
+    formatter = ModelRbiFormatter.new(ApplicationRecord, ['ApplicationRecord'])
+    formatter.generate_rbi
```
manhhung741:
```suggestion
    expect(formatter.generate_rbi).to_not raise_error
```

lib/sorbet-rails/model_rbi_formatter.rb
manhhung741: This is a good addition, but does this mean we still generate an (almost) empty file for this, or there can be an expected rbi file we can add to the test?

pje: >  does this mean we still generate an (almost) empty file for this

Yeah, currently it'd produce an `application_record.rbi` file, and it'll contain all the generated stuff we'd normally see in a model rbi. (note that the rake task filters it out, so we wouldn't hit this line from the rake task anyway.)

I was a little torn on what to do about this case. But if a user asks to generate rbis for an abstract class, it seems to me that we should do it! 

I'd vote to add an expected rbi for this case.

pje: OK, cool—I just added a new abstract class `Potion`, and an expected output rbi for it (just generates relations, no column accessors.)

This seemed like it'd be less confusing than having an `expected_application_record.rbi`, because in the real world people will use the rake task, and that should be blacklisting `ApplicationRecord`.

manhhung741: I agree. Thanks for adding the test!

manhhung741:
Thank you very much for the PR! 🎉

Issue #25, opened 2019-06-24
Support Rails version 6.0.0rc, 5.1, etc.


willisplummer:
Some preliminary discussion around this problem can be found in this [slack thread](https://sorbet-ruby.slack.com/archives/CJL5B3CEN/p1561148847039400)

when i run `rake rails_rbi:models` it skips all of the instance methods for attributes defined in the schema

e.g.

```
-- Generate sigs for ServiceProviderUsersServiceClientUser --
Skip method 'created_at' because it is not autogenerated by Rails.
Skip method 'created_at=' because it is not autogenerated by Rails.
Skip method 'hourly_rate_in_cents' because it is not autogenerated by Rails.
Skip method 'hourly_rate_in_cents=' because it is not autogenerated by Rails.
```

but in a console:

```
irb(main):005:0> ServiceProviderUsersServiceClientUser.instance_method(:hourly_rate_in_cents).source_location
=> ["/bundle/gems/activerecord-5.1.7/lib/active_record/attribute_methods/read.rb", 34]

irb(main):004:0> ServiceProviderUsersServiceClientUser.instance_method(:hourly_rate_in_cents).owner   
=> #<#<Class:0x0000563fdf89fd58>:0x0000563fdf89fda8>
```

## Deps

- ruby 2.6.2
- rails-5.1.7
- activerecord-5.1.7
- sorbet-0.4.4253
- sorbet-runtime-0.4.4253
- sorbet-rails-0.1.3

jasnow:
I see this too.

willisplummer:
@jasnow can you post your versions of ruby, rails, and activerecord?

jasnow:
> @jasnow can you post your versions of ruby, rails, and activerecord?

Rails and activerecord are both 6.0.0.rc1
Ruby 2.6.3.

jasnow:
Will try same repo with Ruby 2.5.5.

jasnow:
Still getting the "Skip method" messages:
```
-- Generate sigs for Micropost --
Skip method 'content' because it is not autogenerated by Rails.
Skip method 'content=' because it is not autogenerated by Rails.
Skip method 'created_at' because it is not autogenerated by Rails.
Skip method 'created_at=' because it is not autogenerated by Rails.
Skip method 'updated_at' because it is not autogenerated by Rails.
Skip method 'updated_at=' because it is not autogenerated by Rails.
Skip method 'user_id' because it is not autogenerated by Rails.
[...]

PR #29, opened 2019-06-25
Refactor instance & class methods to separate modules

manhhung741:
- Refactor to put Model instance methods in separate modules
- Refactor to put Model class methods in separate modules

It follows more correctly what Rails does, and allow people to override the sig if they override the method. Because of that, we don't need to check for overridden methods or matching signatures. The code is unreliable and hard to get correct in different versions of Rails.

To ensure it works with other version of Rails, I also set up tests for Rails 5.1.7. I will add 6.0.0rc later, but one thing to note is that it requires a higher version of Ruby.

See #25


.gitignore
```diff
@@ -42,36 +43,38 @@ build-iPhoneSimulator/
...
+.ruby-version
+.ruby-gemset
 
 # unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
 .rvmrc
 
 # Rails app
-/spec/support/rails_app/tmp/*
-/spec/support/rails_app/.bundle
+<<<<<<< HEAD
```
nacin: merge cruft

lib/sorbet-rails/routes_rbi_formatter.rb
```diff
@@ -15,20 +15,21 @@ def section(routes)
   def header(routes)
   end
 
-  def no_routes(routes, filter)
+  def no_routes(routes, filter=nil)
+    puts routes
```
nacin: debug cruft or intentional?

manhhung741: It's intentional, 5.1.7 only pass the routes and not `filter` to this method.

PR #30, opened 2019-06-25
Add ActiveRecord::Associations::CollectionProxy

halogenandtoast:
Add signatures for ActiveRecord::Associations::CollectionProxy and
update generated model definitions for associations to use
CollectionProxy instead of Relation.


manhhung741:
Thank you for the quick PR! The changes look good, but can you rebase and regenerate the  `expected_portion.rbi`?

halogenandtoast:
@manhhung741 rebased and updated

PR #31, opened 2019-06-25
Script to run tests

manhhung741:
Add scripts to make it easy to run tests & reset test data

### DIFF
--- a/spec/bin/reset_test_data.sh
+++ b/spec/bin/reset_test_data.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# A simple script to reset test data for all rails versions
+# A simple script to reset test data for all Rails versions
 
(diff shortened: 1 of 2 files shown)


spec/bin/run_all_specs.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# A simple script to run all specs for all Rails version
```
nacin:
```suggestion
# A simple script to run all specs for all Rails versions
```

spec/bin/reset_test_data.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# A simple script to reset test data for all rails versions
```
nacin:
```suggestion
# A simple script to reset test data for all Rails versions
```

PR #46, opened 2019-06-26
Change specs to key off Rails branches, not patch versions

nacin:
* Moves `5.1.7` to `v5.1`, etc.
 * The `v` is to make it easier to glob the non-`rails_shared` directories
 * Adds `.codecov.yml` config to make sure it can follow symlinks on the codecov.io dashboard


nacin:
The exact thing this is trying to fix is browsing to a file that is within a directory symlink:
 * https://codecov.io/gh/chanzuckerberg/sorbet-rails/src/904788461eb2483f951e71d89a8e158b596eab6d/spec/support/5.1.7/app/models/wand.rb

Code coverage is run; the dashboard just can't find the file.

The error is:

> `Is the path spec/support/5.1.7/app/models/wand.rb` a relative path in your repository? 
> [Learn how to apply path fixes in your configuration](https://docs.codecov.io/docs/fixing-paths).

.codecov.yml
```diff
@@ -0,0 +1,4 @@
+fixes:
+ # Ensure Codecov can follow files within symlinked directories
```
manhhung741: There are a few symlink files. Do we need to list them as well?

nacin: It's able to follow symlinked files ([it sees routes.rb fine](https://codecov.io/gh/chanzuckerberg/sorbet-rails/src/57b8621503f45920a3fcfb60be70416bb9f4c04b/spec/support/5.1.7/config/routes.rb)), just not files within symlinked directories. It's also not entirely clear that this will fix it, but it _should_.

Gemfile
```diff
@@ -13,10 +13,10 @@ rails =
   case rails_version
   when "master"
     {github: "rails/rails"}
-  when "default"
-    ">= 5.2.3"
+  when "5.1"
+    "~> 5.1.7"
```
manhhung741:
```suggestion
    "~> 5.1.7"
  when "5.2"
    "~> 5.2.3"
```

PR #47, opened 2019-06-28
Test rake tasks

manhhung741:
Set up the structure for testing rake tasks.
Add tests for routes & models rake tasks.


spec/rake_helper.rb
```diff
@@ -0,0 +1,43 @@
+# For testing rake tasks
+# From: https://github.com/eliotsykes/rails-testing-toolbox/blob/master/tasks.rb
```
manhhung741: I'm asking the author about License and submit a PR to make it MIT license. Otherwise I think this is enough attribution? It's actually a great gist of code!

PR #48, opened 2019-06-28
Add rails v4.2

manhhung741:
yada yada it's actually complex to set up support for different rails versions


lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -267,7 +267,10 @@ module #{name}
   end
 
   def type_for_column_def(column_def)
-    cast_type = ActiveRecord::Base.connection.lookup_cast_type_from_column(column_def)
```
AlexEvanczuk: This all looks good, but I think that things might start to get confusing, particularly as Rails 6 comes out soonish.

It might be easier to have classes such as: `SorbetRails::ActiveRecordAdapters::ActiveRecord4`, that can respond to a method like:

```
# Takes in a ActiveRecord::ConnectionAdapters::Column, outputs a ActiveRecord::Type::Value
def get_type_for_column
end
```

I think it will be slightly easier to keep track of different implementations over time and being able to more easily discern what to support and when.  As a stopgap solution seems great though!

Thanks for fixing!

manhhung741: That's a good idea. I'll note it down for future improvement. Please feel free to send a PR to include that as well -- it seems you're in the middle of it.

AlexEvanczuk: I'll look into it when I'm back from vacation!

PR #57, opened 2019-06-28
ActiveRecord "unscoped" query method

gogiel:
Adds support for non-block `unscoped` method: https://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/unscoped

Unfortunately this method has also a less popular block option that has to be untyped as it can return anything so it has to be untyped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,19 @@ Because Sorbet's initial setup tries to flag files at whichever typecheck level 
 
 It is worth going through the list of files that is ignored and resolve them (and auto upgrade the types of other files; see [initial setup](#initial-setup) above). Usually this will make many more files able to be typechecked.
 
+### `unscoped` with a block
+
+[unscoped method](https://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/unscoped) by default, when no block is provided returns a Relation.
+When a block is provided `unscoped` calls the block and returns its result which type is unknown.
+To avoid that either override `unscoped` definition or in your code replace:
+```ruby
+Model.unscoped do … end
+```
+with
+```ruby
+Model.unscoped.scoping do … end
+```
+
 ## Contributing
 
 Contributions and ideas are welcome! Please see [our contributing guide](CONTRIBUTING.md) and don't hesitate to open an issue or send a pull request to improve the functionality of this gem.
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ It is worth going through the list of files that is ignored and resolve them (an
 
 [unscoped method](https://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/unscoped) by default, when no block is provided returns a Relation.
 When a block is provided `unscoped` calls the block and returns its result which type is unknown.
-To avoid that either override `unscoped` definition or in your code replace:
+The gem opts to define `unscoped` as returning a Relation. To avoid that, either override `unscoped` definition or in your code replace:
 ```ruby
 Model.unscoped do … end
 ```
--- a/lib/sorbet-rails/model_rbi_formatter.rb
+++ b/lib/sorbet-rails/model_rbi_formatter.rb
@@ -342,6 +342,7 @@ class ModelRbiFormatter
         prefix = ""
         prefix = "*" if arg_def[:arg_type] == :rest
         prefix = "**" if arg_def[:arg_type] == :keyrest
+        prefix = "&" if arg_def[:arg_type] == :block
 
         "#{prefix}#{arg_def[:name]}"
       }.join(", ")
--- a/lib/sorbet-rails/model_rbi_formatter.rb
+++ b/lib/sorbet-rails/model_rbi_formatter.rb
@@ -85,7 +85,7 @@ class ModelRbiFormatter
     # rails/activerecord/lib/active_record/querying.rb:21
     @generated_scope_sigs["all"] = { ret: "#{@model_class.name}::Relation" }
     @generated_scope_sigs["unscoped"] = {
-      ret: "T.untyped",
+      ret: "#{@model_class.name}::Relation",
       args: [
         { name: :block, arg_type: :block, value_type: 'T.nilable(T.proc.void)' },
       ]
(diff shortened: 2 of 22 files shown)


manhhung741:
The tests for Rails 4.2 is failing as well. Please do update the expected data for v4.2 (or simply rerun `./spec/bin/update_test_data.sh`)

lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -84,6 +84,12 @@ def populate_activerecord_querying_methods
     # All is a named scope that most method from ActiveRecord::Querying delegate to
     # rails/activerecord/lib/active_record/querying.rb:21
     @generated_scope_sigs["all"] = { ret: "#{@model_class.name}::Relation" }
+    @generated_scope_sigs["unscoped"] = {
+      ret: "T.untyped",
```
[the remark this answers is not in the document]
manhhung741: Oops, my mistake. I missed the comment about the block. I'm looking into the optional block and [`.scoping`](https://apidock.com/rails/ActiveRecord/Relation/scoping). Sorbet doesn't support multiple signatures for a method, but making the method return `untyped` isn't useful. I think it's more useful for this method to return a relation, and if people want to use the optional block option, they should use `.scoping`.

Eg:
```
Model.unscoped.scoping do ... end
```
We could add this guidance to the "Tips and Tricks" section.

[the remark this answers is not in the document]
manhhung741: This method should also be added to `@generated_class_sigs` because it only exists in `ActiveRecord::Base` class and not in a `ActiveRecord::Relation`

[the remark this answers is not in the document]
manhhung741: Edit: Never mind, I was wrong, again

gogiel:
Updated. Rails 4.2 specs didn't work out of the box for me and that's why i missed them. I had to `gem install bundler -v '1.17.3'`

manhhung741:
Thanks for fixing the spec! Did you see my suggested change on returning "Model::Relation" instead of "T.untyped"? I still think we should do that.

gogiel:
Sorry, I definitely needed a ☕️

I also found and fixed a problem with the current way definitions with blocks are defined. They were missing `&` character.

```ruby
  sig { params(block: T.nilable(T.proc.void)).returns(T.untyped) }
  def unscoped(block); end
```

didn't work as expected with no arguments, Sorbet was complaining about missing `block` argument.

manhhung741:
I see. Thanks! You may need `block=nil` for it to work when no block is
passed in.

I’ll check out the code locally to test. Sorbet also automatically
generates a def for the method without the block argument. It’s worth
checking if it work out of the box.

Separately I’ll look into setting up sorbet in here to test w it directly.

README.md
```diff
@@ -139,6 +139,19 @@ Because Sorbet's initial setup tries to flag files at whichever typecheck level
 
 It is worth going through the list of files that is ignored and resolve them (and auto upgrade the types of other files; see [initial setup](#initial-setup) above). Usually this will make many more files able to be typechecked.
 
+### `unscoped` with a block
+
+[unscoped method](https://apidock.com/rails/ActiveRecord/Scoping/Default/ClassMethods/unscoped) by default, when no block is provided returns a Relation.
+When a block is provided `unscoped` calls the block and returns its result which type is unknown.
+To avoid that either override `unscoped` definition or in your code replace:
```
manhhung741:
```suggestion
The gem opt to type `unscoped` as returning a Relation. To avoid that, either override `unscoped` definition or in your code replace:
```

lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -336,6 +342,7 @@ def generate_method_sig(method_name, generated_method_def)
         prefix = ""
         prefix = "*" if arg_def[:arg_type] == :rest
         prefix = "**" if arg_def[:arg_type] == :keyrest
+        prefix = "&" if arg_def[:arg_type] == :block
```
manhhung741: Oh, this turns out to be a bigger bug that I thought. Thanks for fixing!

manhhung741:
Could you check the revised message in README? I gave a suggestion to make it clearer. Once you've updated it, I'll merge the PR!

spec/test_data/v4.2/expected_spell_book.rbi
```diff
@@ -72,73 +72,76 @@ module SpellBook::ModelRelationShared
...
-  def preload(*args, block); end
+  def preload(*args, &block); end
 
   sig { params(args: T.untyped, block: T.nilable(T.proc.void)).returns(SpellBook::Relation) }
-  def eager_load(*args, block); end
+  def eager_load(*args, &block); end
 
   sig { params(args: T.untyped, block: T.nilable(T.proc.void)).returns(SpellBook::Relation) }
-  def includes(*args, block); end
+  def includes(*args, &block); end
```
manhhung741: This seems to work on [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Aclass%20Site%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params(a%3A%20T.untyped%2C%20block%3A%20T.nilable(T.proc.void)).returns(String)%20%7D%0A%20%20def%20self.where(*a%2C%20block)%0A%20%20%20%20%22String%22%0A%20%20end%0Aend%0A%0AT.reveal_type(Site.where(id%3A%201))%0AT.reveal_type(Site.where(id%3A%201)%20%7B%20'23'%20%7D%20))
```
./sorbet-test.rb:4: Revealed type: Site::Relation https://srb.help/7014
     4 |T.reveal_type(Site.includes(:id, :name))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  From:
    ./sorbet-test.rb:4:
     4 |T.reveal_type(Site.includes(:id, :name))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^
```

But not when it's a standalone function on [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params(block%3A%20T.nilable(T.proc.void)).returns(String)%20%7D%0Adef%20with_block(block)%0A%20%20'asdf'%0Aend%0A%0AT.reveal_type(with_block())%0AT.reveal_type(with_block%20%7B%20'asd'%20%7D))

It looks like a bug. I'll report it to the sorbet team.

manhhung741: ^ `block` actually works when it's a class method.
However, I still think the fix to use `&block` is better & more consistent.

PR #58, opened 2019-06-29
Use `srb rbi suggest-typed`

manhhung741:
This is a lot more friendly. Under the hood, it is the same

https://github.com/sorbet/sorbet/blob/7e18aed84cdf7ca533eb26aa5cdc2e43aa604b4c/main/options/options.cc#L709

From sorbet documentation 
```
    suggest-typed       Put the highest `typed:` sigil in each of your files
```


manhhung741:
Thanks @jasnow for the suggestion!

PR #65, opened 2019-07-01
Update Readme, section about unscoped

manhhung741:
Revise the writing a little bit.


jasnow:
Do you need to clarify that 2.7.0-preview1 is not included in "later"?

Issue #15, opened 2019-06-20
`reflection.class_name` is sometimes scoped under some namespace and causes constant resolution errors


manhhung741:
We should use `"::#{reflection.klass.name}"` instead of `reflection.class_name` to avoid the problem

https://github.com/chanzuckerberg/sorbet-rails/blob/6fcf542e663ebe015c50dc67f8e4338da020d620/lib/sorbet_rails/models_rbi_formatter.rb#L109

XrXr:
I think I exaggerated this problem in my original message. I must've ran into the error after making modifications. Sorry! The problem is real, but it does not lead to constant resolution errors.

I made an example app to demonstrate the problem. It's nothing fancy, basically just `rails new` and `rails generate model blog/component`

There are two models in the app, Article and Comment, both of which are under the `Blog` namespace. In this case, `reflection.class_name` returns `"Article"` for the `belongs_to :article` association on `Blog::Comment`. sorbet-rails sees and generates T.untyped for the association getter setter, when it could've generated `::Blog::Article` if it used `"::#{reflection.klass.name}"`

https://github.com/XrXr/sorbet-rails-bug-repro/commit/207d3d83a4ab0c65ca2f3656eb268f34cb6ceeea

manhhung741:
Thank you! I'll look at the repo and try to fix this. Looks like a simple fix.

PR #68, opened 2019-07-02
Fixes for sorbet 0.4.4366

manhhung741:
Fix model rbi generation to work with sorbet 0.4.4366

    Because sorbet generates hidden-definition.rbi file that contains the
    following types of methods, they need to be put at the class leve
    directly for sorbet to recognize the signature.
    - Put associations in a model class directly
    - Put enum-related methods in a model class directly
    Note: we keep column methods in a separate modules because we haven't
    supported serializer or attribute API. This allows people to override
    the generated RBI

    Fix using reflection.klass.name instead of reflection.class_name.
    reflection.class_name is a user-input value when they define an
    association, whereas reflection.klass.name reflects the ruby class that
    Rails thinks the association should be
    This also fix https://github.com/chanzuckerberg/sorbet-rails/issues/15

    Use ActiveRecord_Relation & ActiveRecord_Associations_CollectionPolicy
    to match what Sorbet generates in the hidden-definition file.

    Remove SchemaMetadata from blacklist because we need the generation code
    to define `Elem` for the class. (Every subclass of ActiveRecord::Base
    need to redefine this regardless)
   
    Remove .soft

I tested all the changes locally with a Rails codebase to make sure the generated code works with sorbet, and it works out of the box after set up (ie, people don't have to manually fix things).

### DIFF
--- a/lib/sorbet-rails/model_rbi_formatter.rb
+++ b/lib/sorbet-rails/model_rbi_formatter.rb
@@ -40,12 +40,17 @@ class ModelRbiFormatter
     end
     @buffer << draw_module_or_class_footer
 
-    @buffer << draw_module_header("#{@model_class.name}::#{MODEL_CLASS_MODULE_SUFFIX}")
+    # TODO Enum methods need to be defined under the class definition because sorbet generates them
+    # in the hidden-definition.rbi
+    # When this issue is resolved, they might go away by running `srb rbi hidden-definitions`
+    # This is a sure way to make it work though.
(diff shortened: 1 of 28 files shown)


jaredbeck:
Thank you @manhhung741 this seems to have fixed a crash during `srb init` (https://github.com/sorbet/sorbet/issues/1099)

manhhung741:
You’re welcome :-) I’ve released sorbet-rails 0.4.0 that includes all the
changes that should make the gem work much more nicely with sorbet.

PR #73, opened 2019-07-04
Add test for srb tc & srb run time

manhhung741:
Set up the structure to test sorbet against generated rbi.
It supports multiple version of sorbet. We want to maintain 2 version:
- 1 known working version
- 1 master version 
IDK if we should allow the tests against sorbet master to fail. It may remove noise but we may likely forget to check their statuses.

The test examples themselves are in `sorbet_test_cases.rb`. I actually caught an issue with Rails v6.0.0 that the sig for a method change. Yay, tests!

We perform 2 kind of tests:
- `srb tc` and expect the output from T.reveal_type to match output
- load the file & expect the T.assert_type! to not raise error.

Now we have 50 builds per test. Whelp!


Gemfile
```diff
@@ -38,3 +38,14 @@ else
   gem "bundler", ">= 2.0"
   gem "sqlite3", "~> 1.4.1"
 end
+
+sorbet_version = ENV["SORBET_VERSION"]
```
nacin: I don't think `SORBET_VERSION` should be defined to a specific version in Travis. That will result in way too much churn. Instead I'd suggest that we have two versions of this: Sorbet from Rubygems (so, latest release) and sorbet/sorbet's master branch directly from GitHub.

However, at least for now, these are exactly the same: Each commit to sorbet/sorbet appears to be deployed directly to rubygems. (The current version is 0.4.4389, which lines up with 4,389 commits to master.) So for now, we probably just need to test against Sorbet as coming from Rubygems, as we were doing already, but we should probably expect their release process to change in the future.

spec/sorbet_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  before(:each) do
+    Rake::Task['rails_rbi:routes'].invoke
+    Rake::Task['rails_rbi:models'].invoke
+  end
+
+  it 'returns expected sorbet tc result' do
+    # invoke the rake tasks
+    cmd = "cd #{Rails.root.to_path}\; srb tc"
+    stdout, stderr, status = Open3.capture3(cmd)
```
nacin: This should instead be something like `Open3.capture3('srb', 'tc')`. Entire commands should never be passed to `Open3` as a single string for security reasons. An additional argument of `:chdir => Rails.root.to_path` can handle the directory change as well.

Gemfile
nacin: I guess my point is, for as long as the latest version of the Sorbet gem is going to be their HEAD commit, we probably should just always test on that version, rather than also testing on a known earlier version.

PR #76, opened 2019-07-05
Fix using `sorbet-rails` with engines

masylum:
## Problem

Right now, when you run the rake task to populate models rbis, it fails to populate models in your engines.
At [https://factorialhr.com](Factorial) we use engines heavily to isolate the different domains of our application.

## Solution

In Rails there is a way to force eager load for all namespaces as indicated in this thread: https://stackoverflow.com/questions/19839847/rails-force-models-to-eager-load


manhhung741:
Thanks!

The change looks good to me. Can you add a new test that cover an engine
model that isn’t loaded before the change but will be after the change?

I wonder if this loads “ActiveRecord::InternalMetadata” and
“ActiveRecord::SchemaMigration”. If it does, we can remove the whitelists
too!

lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -21,7 +21,7 @@ namespace :rails_rbi do
   desc "Generate rbi for rails models. Pass models name to regenrate rbi for only the given models."
   task models: :environment do |t, args|
     # need to eager load to see all models
-    Rails.application.eager_load!
+    Rails.configuration.eager_load_namespaces.each(&:eager_load!)
```
manhhung741: Looking at the stackoverflow link & it looks like some people running into issue calling `eager_load!`. I think it might be safer to use `try`
```suggestion
    Rails.configuration.eager_load_namespaces.each { |s| s.try(:eager_load!) }
```

manhhung741:
@masylum Hey, I think this will be a good addition for 0.4.0 so I added spec to the PR

PR #78, opened 2019-07-06
Sorbet init test

manhhung741:
Add a test config that test the whole flow:
- installing all gems
- srb init
- run `sorbet-rails` rake tasks
- run `srb tc` and other tests.


.travis.yml
```diff
@@ -19,17 +20,27 @@ rvm:
...
-    env: RAILS_VERSION=5.2 # SORBET_VERSION=0.4.4386
+    env: RAILS_VERSION=5.2 SORBET_VERSION=0.4.4386
   - rvm: 2.6
-    env: RAILS_VERSION=5.2 # SORBET_VERSION=0.4.4386
+    env: RAILS_VERSION=5.2 SORBET_VERSION=0.4.4386
   - rvm: ruby-head
-    env: RAILS_VERSION=5.2 # SORBET_VERSION=0.4.4386
+    env: RAILS_VERSION=5.2 SORBET_VERSION=0.4.4386
+  # only need to have 1 test with srb init
+  - rvm: 2.3
```
nacin: Given the intention, would this be better as an `include:`, rather than adding to `matrix:` and excluding all but one?

manhhung741: 👍

PR #85, opened 2019-07-15
Add vim swap files to .gitignore

palasha:
As title. These get annoying to ignore real fast. I've added these to many repos in the past and never had any issues.


gogiel:
Locally you can use your own, global `.gitignore`. https://stackoverflow.com/a/5724484

These files have nothing to do with sorbet-rails.

jgable:
I agree with the @gogiel that this should be handled with his own global .gitignore. However, if @palasha is finding it annoying to keep them out of PR’s then some newcomer to contributing will also inadvertently include them and this will prevent that.

TL;DR: I'd rather explicitly exclude them like this so we don't have to fight them in PR's in the future.

manhhung741:
Thanks for chiming in! I'm going to merge this change because it can prevent bad files checked in.

PR #86, opened 2019-07-15
Regenerate descendants of specific models

palasha:
When using STI in rails with a large number of descendant models, it's frustrating when I update the parent class because I have two options:

1. Re-run the entire `rake rails:models_rbi` script, which is slow.
2. Enumerate all child models manually and make a really long command like `rake rails:models_rbi[Foo,Bar,Bax,Asdf,Etc]`

This changes the default to *include* descendants of a model when its specified as a specific argument. Then I only need to do `rake rails:models_rbi[Parent]` to regenerate all children.

I think this makes sense because I can't really imagine a scenario where the parent changes but we would not want to also be safe and regenerate the children.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ This Rake task generates RBI files for all models in the Rails application (all 
 ```sh
 ❯ rake rails_rbi:models
 ```
-You can also regenerate RBI files for specific models:
+You can also regenerate RBI files for specific models. To accommodate for STI, this will generate rbi for all the subclasses of the models included.
 ```sh
 ❯ rake rails_rbi:models[ModelName,AnotherOne,...]
 ```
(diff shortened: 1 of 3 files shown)


palasha:
Interesting, I wonder why this would pass locally but not remotely. Shouldn't the test script have regenerate the rbi's for Wizard model as well?

manhhung741:
I agree this will be a good change!

It looks like you need to run `rake db:migrate` in each rails version folder. I keep db/schema.rb separate because each database will have a different db, and we need `rake db:migrate` each folder separately to apply the change to the schema.

Since it's only test db, we probably can simplify this flow by using the same db/schema.rb and always resetting db somehow.

manhhung741:
Hey @palasha do you have time to update the testcases for this PR?

manhhung741:
@palasha I updated the diff for you ^^

Issue #87, opened 2019-07-17
Generate a helpers.rbi


connorshea:
Rails helpers are generated with no parent, so Sorbet thinks they don't include `Kernel`, which means something like this will cause the typechecker to fail, despite being perfectly valid:

```ruby
# typed: true
module ApplicationHelper
  def example
    puts "test"
  end
end
```

If a Rails app has helper modules, a `helpers.rbi` should be generated like so:

```ruby
# typed: strong
module ApplicationHelper
  include Kernel
end

module GamesHelper
  include Kernel
end

module SettingsHelper
  include Kernel
end
```

I'm not sure if this is quite the correct fix, but I think it makes sense. It'd also be a base to build from for better helper support in the gem.

manhhung741:
Sounds like a good idea. If you'd like to send a PR to include this to the generation code, please feel free to do so. 

Are there more things you'd like to add to the helpers?

connorshea:
ActionView methods, mostly. I'm still trying to figure out what exactly the included actionview classes/modules are, though.

connorshea:
you can get a list of Helper modules with `ApplicationController.modules_for_helpers([:all])`
```
irb(main):011:0> ApplicationController.modules_for_helpers([:all])
=> [ApplicationHelper, GamesHelper, GenresHelper, HomeHelper, PlatformsHelper, SettingsHelper, DeviseHelper]
```

PR #88, opened 2019-07-17
Add Helpers generation task.

connorshea:
Implements #87.

Right now this is very basic and only includes Kernel to allow for basic methods like `puts` or `raise`, in the future we can update it to include more modules from ActionView and elsewhere.

I also still need to update the tests for this change.

This is the kind of `rbi` it generates for [my Rails app](https://github.com/connorshea/VideoGameList):

```ruby
# typed: strong

module ApplicationHelper
  include Kernel
end

module GamesHelper
  include Kernel
end

module GenresHelper
  include Kernel
end

module HomeHelper
  include Kernel
end

module PlatformsHelper
  include Kernel
end

module SettingsHelper
  include Kernel
end

module DeviseHelper
  include Kernel
end
```


connorshea:
As far as I can tell, the failures are caused by a change in the upstream Sorbet project?

As for the hack that fixes my issue, I wasn't able to get using a version check to work, so I just ended up trying both and setting the helpers array dependent on which one worked.

manhhung741:
Yep, the test with a previous version of sorbet still passed. I'm glad I had it there.
I messaged the sorbet team about the new error:
https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1563377281497000

manhhung741:
Can you add a test to test the rake task itself? It's worth testing the loading logic. ApplicationController is introduced in Rails 5.0 (or 5.1?), maybe it's a different behavior between different version of Rails.

manhhung741:
I had a PR out to fix the tests for latest version of sorbet.

connorshea:
Are there currently any tests for the other rake tests? I’ll try to get to that tonight.

manhhung741:
Yes, there are! Checkout spec/rake_rails_rbi_models_spec.rb for example

connorshea:
I'm not sure how to get this working, honestly. Regenerating the expectations causes them to return with no helpers, for some reason :/

manhhung741:
:’( I’ll see what I can do to help.

manhhung741:
It might still not load helpers correctly.

manhhung741:
I pushed a fix to your branch that change `ApplicationController` to subclass `ActionController::Base` instead of `ActionController::API`. 
You may need to undo unrelated expected data changes because I fixed the conflicts with sorbet upstream

connorshea:
Yup, looks like it needs to have the expectations updated

connorshea:
@manhhung741 this should be ready to merge now 👍

manhhung741:
Thank you!

PR #91, opened 2019-07-18
Allow non-integer value types for enum.

xaviershay:
[Documentation states](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) these should be integers, but any object works
(most commonly: String).

This PR isn't complete:

* I'm not sure the best way to spec this. That's probably not hard, I just don't want to sink time into figuring out before confirming this is a change we want :) Presume I add another column to one of the fixtures? Or I could maybe change one of the existing, since their are two `enum` examples already.
  * Note to self: relevant info in [contributing guidelines](https://github.com/chanzuckerberg/sorbet-rails/blob/master/CONTRIBUTING.md).
* There's an argument to _not_ merge this, since this is effectively enabling typing of (for what I can tell) an undocumented use case. But I believe using strings here is relatively common: it's suggested in [multiple](https://stackoverflow.com/questions/32938729/how-to-store-enum-as-string-to-database-in-rails/32939934) [internet](https://stackoverflow.com/questions/24105813/possibility-of-mapping-enum-values-to-string-type-instead-of-integer) [locations](https://sipsandbits.com/2018/04/30/using-database-native-enums-with-rails/) (top 3 google results, didn't look further). If we choose to not merge though, I think there needs to be a documented workaround in this issue to help more people adopt sorbet.


lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -215,7 +215,20 @@ def populate_collection_assoc_getter_setter(assoc_name, reflection)
...
-      @generated_class_sigs["#{enum_name.pluralize}"] = { ret: "T::Hash[T.any(String, Symbol), Integer]"}
+      # Rails documents these values as being integers, but in practice any
+      # type is allowed.
+      value_types = enum_hash.values.map {|x| x.class.to_s }.uniq
+      value_type = if value_types.length > 1
+        "T.any(%s)" % value_types.join(', ')
+      elsif value_types.length == 1
+        # This case not strictly required, it's just neater.
+        value_types.join(', ')
+      else
```
manhhung741: This would mean no `value_types` is defined. When does that happen?

lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -215,7 +215,20 @@ def populate_collection_assoc_getter_setter(assoc_name, reflection)
 
   def populate_generated_enum_methods
     @model_class.defined_enums.each do |enum_name, enum_hash|
-      @generated_class_sigs["#{enum_name.pluralize}"] = { ret: "T::Hash[T.any(String, Symbol), Integer]"}
+      # Rails documents these values as being integers, but in practice any
+      # type is allowed.
+      value_types = enum_hash.values.map {|x| x.class.to_s }.uniq
```
manhhung741:
```suggestion
      value_types = enum_hash.values.map {|x| x.class.name }.uniq
```

manhhung741: Can we use `value_types` for the sig at line 228?

```
        @generated_class_sigs["#{enum_val}"] = {
          args: [ name: :args, arg_type: :rest, value_type: 'T.untyped' ],
          ret: @model_relation_class_name,
        }
```
(I think for this method, symbol & string are always acceptable as input, besides the value_types)

lib/sorbet-rails/model_rbi_formatter.rb
xaviershay: Mostly defensive coding. `enum incomplete: {}` is syntactically correct and might be part of a WIP.

I can see an argument for "raising an error" somehow but not sure the right place: I don't think this generation process should be interrupted. Could use `T.untyped`? Or maybe there's a sorbet equivalent to `T.error`?

xaviershay:
thanks for feedback! Will add tests cases.

manhhung741:
FYI I've merged some changes that refactor this code into its own file (plugin). The code now is in this file:
https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/model_plugins/active_record_enum.rb

xaviershay:
nice :tada:

manhhung741:
Oops, sorry for the confusion. I meant the code related to enum is moved &
when you rebase you’d need to add the code there. I didn’t mean I added
changes from your PR there.

xaviershay:
_clicks reopen_

manhhung741:
I believe #145 handle this but more up-to-date with `sorbet-rails` codebase.

manhhung741:
@xaviershay I'll close the issue because #145 includes this fix. Please let me know if you want to add changes on top of it.

PR #93, opened 2019-07-21
Include ActionView::Helpers in helpers.rbi

connorshea:
`ActionView::Helpers` provides methods like `link_to` in Helpers, e.g. I have a helper like this:

```ruby
# typed: true
module GamesHelper
  def sort_dropdown_link(sort, humanized_name)
    link_to(humanized_name, request.params.merge(order_by: nil), class: "dropdown-item#{params[:order_by].nil? ? ' has-text-weight-bold' : ''}")
  end
end
```

Previously, the `link_to` and `request` methods were undefined according to Sorbet. With this change, this code no longer throws Sorbet errors :)

I had hoped to also figure out how to include the URL methods (e.g. `user_url`) in Helpers, but I think the way sorbet-rails currently implements those methods is somewhat flawed (or perhaps I'm misunderstanding how it actually works in Rails?), since I haven't been able to figure out a way to include those methods as part of the helper modules. If I'm missing something, I'd appreciate any help figuring this out :)

### DIFF
--- a/lib/sorbet-rails/helper_rbi_formatter.rb
+++ b/lib/sorbet-rails/helper_rbi_formatter.rb
@@ -23,6 +23,7 @@ class HelperRbiFormatter
       rbi += <<~RBI
         module #{helper}
           include Kernel
+          include ActionView::Helpers
         end
 
       RBI
--- a/spec/test_data/v4.2/expected_helpers.rbi
+++ b/spec/test_data/v4.2/expected_helpers.rbi
@@ -4,13 +4,16 @@
 
 module BarHelper
   include Kernel
+  include ActionView::Helpers
 end
 
 module BazHelper
   include Kernel
+  include ActionView::Helpers
 end
 
 module FooHelper
   include Kernel
+  include ActionView::Helpers
 end
 
--- a/spec/test_data/v5.0/expected_helpers.rbi
+++ b/spec/test_data/v5.0/expected_helpers.rbi
@@ -4,13 +4,16 @@
 
 module BarHelper
   include Kernel
+  include ActionView::Helpers
 end
 
 module BazHelper
   include Kernel
+  include ActionView::Helpers
 end
 
 module FooHelper
   include Kernel
+  include ActionView::Helpers
 end
 
--- a/spec/test_data/v5.1/expected_helpers.rbi
+++ b/spec/test_data/v5.1/expected_helpers.rbi
@@ -4,13 +4,16 @@
 
 module BarHelper
   include Kernel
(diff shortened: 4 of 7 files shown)


manhhung741:
Thanks Connor. I’ll reply the bit about routes url:

The URL helper is a module created in the fly in Rails. There isn’t a
consistent name for it.

Yes, the way sorbet-rails generate is not ideal. Currently the methods are
put into ApplicationController directly. My original usecase was to make
them available in ApplicationController.

Now how do we deal with this? We can “trick” sorbet the following way:
- Make up a module that contains all the url helper methods. This is the
trick I use for a lot of the models modules. Say it’s GeneratedUrlHelper.
- Make ApplicationController include this module in an rbi file.
- Make helper modules include this module in an rbi file.

How does this trick work? In static check, sorbet doesn’t validate the
existence of the module. It’ll happily use the fake module defined in the
classes/modules for static check. During run time check, the methods should
be available. Even though it’s under another module, it is still fine.

Ie:
```
module GeneratedUrlHelper
# _url and _path put here
end

class ApplicationController
  include GeneratedUrlHelper
end

module FooHelper
  include GeneratedUrlHelper
end
```

Let me know if this makes sense to you :-)

connorshea:
Ooh, clever

I'll try it. 👍

connorshea:
For now I think we can probably just merge this (assuming you're fine with the current code), we can do the url/path methods later.

manhhung741:
Ok :-)

jaredbeck:
> .. we can do the url/path methods later.

Did this happen? I'm still getting:

```
app/helpers/redacted/redacted_helper.rb:9: 
Method redacted_path does not exist on Redacted::RedactedHelper 
https://srb.help/7003
     9 |        redacted_path(redacted: { redacted: redacted })
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 1
```

Do you want me to create an issue for this?

manhhung741:
It might’ve just happened with v0.5.2. Which version of the gem are you
using? Would you mind upgrading to 0.5.2 and try again?

If the issue persists, please feel free to file an issue!

jaredbeck:
> Would you mind upgrading to 0.5.2 and try again?

After upgrading from 0.5.1 to 0.5.2, I no longer get the error 7003 I showed above. Thanks!

manhhung741:
Forward the thank to @connorshea :P

PR #99, opened 2019-07-26
Rails application type generator

### LINKED ISSUE #98: Type of Rails.application should be YourApp::Application
Moved from: https://github.com/sorbet/sorbet-typed/issues/83

**Describe the bug:**
In a Rails application the type of [Rails.application](https://github.com/sorbet/sorbet-typed/blob/d17bc9ee37629ed824f78d3a2f92e99a8b4e8709/lib/railties/all/railties.rbi#L5) is actually a subclass of `Rails::Application` that is generated when you type `rails new`. For example, `rails new blog`, `Rails.application` will create a class called `Blog::Application` defined in `config/application.rb`.

I am willing to make an attempt at fixing this if you think it's a valid change - let me know then I'll send a PR.

**Steps to reproduce:**

The impact of this is that any custom methods you define in `application.rb` on `Blog::Application` will not be visible to the type checker. If I have `config/application.rb` that looks like this:

```ruby
# typed: true
require_relative 'boot'
require 'rails'

# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)

module Blog
  class Application < Rails::Application
    # Initialize configuration defaults for originally generated Rails version.
    config.load_defaults 5.2

    def name
      "Brad's awesome blog"
    end
  end
end
```

then I can't call `Rails.application.name` in a type-safe way.

**Versions:**

```
$ ruby --version
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]

$ rails --version
Rails 5.2.1

$ srb --version
Sorbet typechecker 0.4.4429 git 51504253c985d0a967d3df6a39ac44b25db2c481 built on 2019-07-12 00:02:11 GMT with debug symbols
```

bradleybuda:
A first attempt to fix #98. The code is straightforward, but it has a (fatal) issue - the `Rails#application` type I'm generating in `rake rbi:application` conflicts with the `Rails#application` [type defined in sorbet-typed](https://github.com/sorbet/sorbet-typed/blob/master/lib/railties/all/railties.rbi#L4), and I don't know what the resolution rules in sorbet are when two different `.rbi` files try to define different signatures on the same method - from playing around, I think the "last one in" wins, but it's not clear if this is well-defined or how we'd exploit it.

Some possible fixes:
1. Remove `Rails.application` types from `sorbet-typed/railties`. This feels wrong because it makes sorbet less useful for Rails users who are not `sorbet-rails` users, and creates a backwards maintenance dependency from `sorbet-typed` to `sorbet-rails`.
2. Update my generator to delete the railties type from the vendored `sorbet/rbi/sorbet-typed/lib/railties/all/railties.rbi` file. This is potentially fragile but it doesn't seem crazy to "manage" vendored types.
3. Reach out to the sorbet team to find the "right way" for types to override other types (feels like a lot of complexity but maybe there are other use-cases for this?)

Thoughts? I'm going to proceed with option 2 for now.


manhhung741:
The conflict sig is a challenging issue. I think (2) is not the right approach because the file can contain other signatures as well. Removing the file will remove more than the application signature. 
There is a (4) option: instruct users to remove the `self.application` sig from the railtie.rb file manually when using this rake task, but that's prone to error when people update the sorbet-typed library.

Let's post in sorbet-ruby slack and see what the sorbet team suggest.

connorshea:
Here's the Parlour version of the RBI generator:

```ruby
root.create_class(Rails.application.class.name, superclass: 'Rails::Application')

rails_module = root.create_module('Rails')
rails_module.create_method(
  'application',
  return_type: Rails.application.class.name,
  class_method: true
)
```

manhhung741:
Closing because this PR has been outdated for a long time. @bradleybuda If you still want to make the change, please send another PR based on the newest code

PR #100, opened 2019-07-29
RFC Typecheck code in lib/ with sorbet

manhhung741:
Adding sorbet & signature for most of sorbet-rails classes & methods. This provides some typecheck level to methods in sorbet.

The result looks promising:
```
               types.input.files :        38
  types.input.files.sigil.autoge :         1
   types.input.files.sigil.false :         3
  types.input.files.sigil.strict :        10
  types.input.files.sigil.strong :        19
    types.input.files.sigil.true :         5
       types.input.methods.total :     12519
  types.input.methods.typechecke :        60
         types.input.sends.total :       728
         types.input.sends.typed :       592
                 types.sig.count :      1865
```

Caveat: turns out I don't know how we should provide sig for ActiveRecord::Base & related methods in a way that'll support multiple versions of Rails. Instead, I opt to give a fake "activerecord.rbi" that contains sigs of the methods used. Consequently, some methods ActiveRecordAssoc plugin is not typed because it doesn't make sense to typedef them with `T.untyped` for reflection


connorshea:
FYI you should probably ignore sorbet/ in the gemspec so the packaged gem isn’t as big

manhhung741:
That's a good idea. Thanks! I also need to make it so that we run sorbet in CI as a check.

PR #101, opened 2019-07-30
Update the helper formatter to use Parlour.

connorshea:
And update test expectations to match.


manhhung741:
Looks good to me. I see you talk to Aaron about adding blank line though. I agree it's more consistent with our editor set up. LMK whenever you want to merge it.

connorshea:
I don't think it's a big enough to deal to block merging for now, unless you think it's a problem. We can update parlour whenever we get that fixed.

PR #105, opened 2019-08-02
Update Rails 6 to 6.0.0.rc2 

connorshea:
Also removes `errors.txt` and `hidden.rbi` from the rails 6 spec, I think these were supposed to be removed in https://github.com/chanzuckerberg/sorbet-rails/commit/edc5f801575dd0b195b887cd9b0b0e5ab091b346 but never were.


manhhung741:
Thanks! The error.txt & hidden.rbi were kept because Sorbet wasnt working
for Rails 6.0.0.rc2 and `srb rbi hidden-definition` wouldn’t generate them.
I might have a check somewhere to make it work with 6.0.0.rc2 in the past?
Let me double check.

PR #109, opened 2019-08-05
Add Kaminari plugin.

connorshea:
I'm not sure how exactly you want this to work, but I figured I'd start by opening a PR since I got a working Kaminari plugin done :)


connorshea:
hmmm, I wonder if I can just wrap any uses of `Kaminari` in T.unsafe 🤔

manhhung741:
If you run `srb rbi todo` it'll put Kaminari into todo.rbi. You can also define it in a `rbi` file like I do for ActiveRecord

manhhung741:
Should we add kaminari to a rails sample app (at least 5.2) and add some `sorbet_test_cases` for it?

connorshea:
I'm kind of hesitant to set up tests for plugins like this just because it'll add so much more complexity to the test suite when it's already really complex, and I think a lot of the value of these plugins is making it easy to contribute them.

manhhung741:
I think eventually we'd need a way to test these plugins. Otherwise we'll run into the same problem with `sorbet-typed`. In terms of complexity, I think we might want 1 special app to test gems. At least the code should be smoke tested to work as intended.
Without test, it makes it *harder* to contribute because you don't have confidence that your changes work.

I agree we shouldn't add the gem to the default list of plugins. I'll start a list of available gem plugins people can add to their list.

lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
```diff
@@ -0,0 +1,21 @@
...
+  # Kaminari generates a dynamic `page` method on ActiveRecord relations.
+  # https://github.com/kaminari/kaminari/blob/c5186f5d9b7f23299d115408e62047447fd3189d/kaminari-activerecord/lib/kaminari/activerecord/active_record_model_extension.rb#L15
+  sig { params(root: Parlour::RbiGenerator::Namespace).void.implementation }
+  def generate(root)
+    return unless @model_class.include?(Kaminari::ActiveRecordModelExtension)
+
+    # Get the configured Kaminari page method name, or fall back to 'page' if necessary.
+    page_method = T.unsafe(Kaminari).config.page_method_name || 'page'
+
+    ar_querying_rbi = root.create_module(model_relation_shared_module_name)
```
manhhung741: This is fine but a more sustainable approach is to add a model-scoped Kaminari-plugin. Is `page` the only method added by the gem?

lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
```diff
@@ -0,0 +1,24 @@
+# typed: true
+class KaminariPlugin < SorbetRails::ModelPlugins::Base
+  # Kaminari generates a dynamic `page` method on ActiveRecord relations.
+  # https://github.com/kaminari/kaminari/blob/c5186f5d9b7f23299d115408e62047447fd3189d/kaminari-activerecord/lib/kaminari/activerecord/active_record_model_extension.rb#L15
+  sig { implementation.params(root: Parlour::RbiGenerator::Namespace).void }
+  def generate(root)
+    return unless @model_class.include?(Kaminari::ActiveRecordModelExtension)
+
+    ar_querying_rbi = root.create_module(model_relation_shared_module_name)
+    ar_querying_rbi.create_include('Kaminari::ActiveRecordModelExtension')
```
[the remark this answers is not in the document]
connorshea: This is what it used to look like, but the current version just adds it directly to `Wizard::ModelRelationShared`

lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
connorshea: I think it's the only method, yeah.

connorshea:
With the new test system it shouldn't be too hard to add an app specifically for plugins, it'd just require adding a check to the generator for a `PLUGINS_ENABLED` environment variable, or something like that.

lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
manhhung741: I see, this is fine then 👍

manhhung741:
Since the plugin is simple, and not included by default, I'll just merge it. We can set up a test app when there are more gem plugins introduced.

lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
ghiculescu: It actually adds a few other methods by way of including `Kaminari::PageScopeMethods` and `Kaminari::ActiveRecordRelationMethods`. For example, this is valid Kaminari: `User.page(1).per(10)` but currently only the `page` call is recognised.

ghiculescu: What do you think about adding an `includes Kaminari::PageScopeMethods` to the model as part of this plugin?

connorshea: That sounds fine to me :)

ghiculescu: https://github.com/chanzuckerberg/sorbet-rails/pull/228

PR #110, opened 2019-08-05
 Got rid of toplevel messages

jasnow:
Got rid of "warning: class variable access from toplevel" messages.


lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -67,7 +63,8 @@ namespace :rails_rbi do
   end
 
   def copy_bundled_rbi
-    bundled_rbi_path = File.join(@@sorbet_rails_rake_dir, "..", "rbi", ".")
+    sorbet_rails_rake_dir = "#{Dir.pwd}/lib/sorbet-rails/tasks"
```
manhhung741: I'm not sure this work when people install the gem & run it from rake. `Dir.pwd` will refer to the location they run the rake task, not the root directory of the gem.

jasnow: How about: sorbet_rails_rake_dir = "#{Gem::Specification.find_by_name("sorbet-rails").gem_dir}/lib/sorbet-rails/tasks"

jasnow: Let's try.

lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -3,10 +3,6 @@ require("sorbet-rails/routes_rbi_formatter")
 require("sorbet-rails/helper_rbi_formatter")
 require("sorbet-rails/utils")
 
-# this is ugly but it's a way to get the current directory of this script
-# maybe someone coming along will know a better way
-@@sorbet_rails_rake_dir = File.dirname(__FILE__)
```
palasha: Could a simpler way to resolve this be to define an inline class with a constant?

eg:
```
class Foo
  RAKE_DIR = ...
end
```

And then reference `Foo::RAKE_DIR` at the callsite?

jasnow: I will close this PR.

PR #112, opened 2019-08-08
Added err and err.class to rescue puts statement to get more information

jasnow:
Added err and err.class to rescue puts statement to get more information.


jasnow:
Here are the resulting messages after do this PR:
```
ActiveRecord::StatementInvalid: Note: Unable to create new instance of ActiveStorage::Attachment
ActiveRecord::StatementInvalid: Note: Unable to create new instance of ActiveStorage::Blob
ActiveRecord::StatementInvalid: Note: Unable to create new instance of ActiveStorage::Attachment
ActiveRecord::StatementInvalid: Note: Unable to create new instance of ActiveStorage::Blob
```

jasnow:
Thanks

PR #113, opened 2019-08-09
Update enum fields to be nilable

rnvarma:
Previously, we were generating the type of enum fields to be not nullable. However, active record supports nilable enum fields.

This updates the type generation to support this behavior


manhhung741:
Congrats on your first PR! @rnvarma

PR #117, opened 2019-08-11
Change the spec/support/ Rails apps to Rails App Templates

connorshea:
This allows regenerating the Rails apps in `spec/support/` from scratch, and centralizes all the custom code/version differences into `spec/generators/rails-template.rb`. It adds Rake tasks that can be used to regenerate the apps (`bundle exec rake update_spec:v6_0`, `bundle exec rake update_spec:v5_2`, etc.)

We _could_ remove the apps entirely from the `spec/support/` directory, but I'm not sure if that's a great idea just because of the potential consistency problems it could create.

Also, recent versions of Rails add the ruby version that generated the app to the `Gemfile`, I should probably figure out how to fix that.

I haven't actually run the tests yet, so I don't know how much they'll need to be changed for this :)


connorshea:
To regenerate the 4.2 app I have to do the following:

- `gem uninstall bundler -v 2.0.2` (and any other versions of 2.x I have installed)
- `rm Gemfile.lock`
- `RAILS_VERSION='4.2' bundle install`
- `RAILS_VERSION='4.2' bundle exec rake spec_update:v4_2`

So right now `bundle exec rake spec_update:all` will work for the other versions, then fail on 4.2.

spec/generators/rails-template.rb
```diff
@@ -0,0 +1,269 @@
...
+def create_initializers
+  initializer "sorbet_rails.rb", <<~RUBY
+    # typed: strict
+    require(Rails.root.join('lib/mythical_rbi_plugin'))
+    SorbetRails::ModelRbiFormatter.register_plugin(MythicalRbiPlugin)
+  RUBY
+end
+
+def create_lib
+  lib "mythical_rbi_plugin.rb", <<~'RUBY'
```
manhhung741:
```suggestion
  lib "mythical_rbi_plugin.rb", <<~RUBY
```

spec/generators/rails-template.rb
```diff
@@ -0,0 +1,269 @@
...
+  RUBY
+end
+
+def add_sorbet_test_files
+  file "typed-override.yaml", <<~YAML
+    true:
+    - ./sorbet_test_cases.rb
+  YAML
+
+  copy_file "./sorbet_test_cases.rb", "sorbet_test_cases.rb"
```
manhhung741: I think this is one instance that creating a symlink makes more sense than copying the file. We may want to add test cases to this file & it should be easy to sync without having to rerun the generation script.

spec/generators/rails-template.rb
```diff
@@ -0,0 +1,269 @@
...
+  create_helpers
+  create_models
+  create_migrations
+  add_sorbet_test_files
+  
+  # Rails 4.2 doesn't have the rails_command method.
+  if ['4.2'].include?(ENV["RAILS_VERSION"])
+    run "bundle exec rake db:migrate"
+  else
+    rails_command "db:migrate"
```
manhhung741: if you want to simplify it, we can always use `bundle exec rake db:migrate`. It works on all versions of Rails.

manhhung741:
First, this is great! Thank you so much. The test set-up makes is a lot clearer with this. What's more, all custom files are in the generation file. It's very easy to know what's customized.

However, I'm still thinking how to make it easy to for iterating on changes.
- How long does it take to re-run the generation for all files?
- Maybe we should add a command to quickly sync `sorbet_test_cases.rb` to all test folders. It might be a file most frequently changed in the setup.
- Maybe we should document how to run just part of the generation script (for adding new migrations or models).

spec/generators/rails-template.rb
connorshea: This is necessary because otherwise the `#{}` will get interpreted here, instead of being inserted into the file.

manhhung741: Oh, that's interesting. Thanks for clarification.

manhhung741:
@connorshea Would you like me to merge this PR and we look into the follow ups later?

connorshea:
@manhhung741 I've resolved some of the feedback that I could fix quickly, but for the remainder we can merge this and fix them afterward :)

manhhung741:
Thank you so much @connorshea Please feel free to address other follow ups. Or we can file them as issues to keep track of them.

PR #118, opened 2019-08-11
Added dalli and redis gems to Gemfile

jasnow:
Added dalli and redis gems to Gemfile because I found 'require "redis"' and 'require "dalli"' in Rails codebase even if not in app.

This also gets rid of the following error messages during sorbet-rails "rake" run.
```
You don't have dalli installed in your application. Please add it to your Gemfile and run bundle install
You don't have dalli installed in your application. Please add it to your Gemfile and run bundle install
The Redis cache store requires the redis gem, version 4.0.1 or later. Please add it to your Gemfile: `gem "redis", "~> 4.0"````


connorshea:
I could be wrong but I don't think this should be necessary? Rails just throws warnings because Sorbet runs every file - including the ones that are normally only used when dalli/redis are enabled - there's not really any negative effect from not having these gems.

manhhung741:
Agree with Connor. I’d be interested to know what files are printing those
messages and can we remove them. That sounds more appropriate to me than
adding the gems that we don’t need.

jasnow:
* actionpack-5.2.3/lib/action_dispatch/middleware/session/mem_cache_store.rb:
   $stderr.puts "You don't have dalli installed in your application.
   Please add it to your Gemfile and run bundle install"

 * activesupport-5.2.3/lib/active_support/cache/mem_cache_store.rb:
   $stderr.puts "You don't have dalli installed in your application.
   Please add it to your Gemfile and run bundle install"

jasnow:
activesupport-5.2.3/lib/active_support/cache/redis_cache_store.rb:  warn "The Redis cache store requires the redis gem, version 4.0.1 or later. Please add it to your Gemfile: `gem \"redis\", \"~> 4.0\"`"

jasnow:
I agree with the "unnecessary" comment. 

If you can selectively ("when they are not enabled") 
filter out the error messages, that would be okay with me too.

jasnow:
After more thought, I see my change as a workaround to 
get rid of the messages so I will close this PR. 
I hope someday I won't need the workaround.

PR #121, opened 2019-08-14
Added DISABLE_DATABASE_ENVIRONMENT_CHECK=1; Upgraded sorbet-rails gem to 0.5.0

jasnow:
1. Add 'DISABLE_DATABASE_ENVIRONMENT_CHECK=1' twice to sorbet-rails/spec/bin/run_spec.sh file.
 2. Then I run sorbet-rails/spec/bin/run_spec.sh file to prove fix worked. As a side effect, the sorbet-rails gem was upgraded to 0.5.0 in each Gemfile.lock file.

The first fix gets rid of these messages:
```
ActiveRecord::NoEnvironmentInSchemaError:

Environment data not found in the schema. To resolve this issue, run:

        bin/rails db:environment:set RAILS_ENV=test
```


Gemfile.lock
```diff
@@ -0,0 +1,186 @@
+PATH
```
manhhung741: We git-ignore Gemfile.lock to support multiple version of ruby & rails. Please remove this file

jasnow: done

jasnow:
I'm going to redo this PR another way.

PR #123, opened 2019-08-15
Convert routes formatter to use Parlour

### LINKED ISSUE #102: Update routes formatter to use Parlour
The routes formatter should be updated like the model formatter to use Parlour, for consistency :)

connorshea:
Fixes #102.

This changes the output a bit, but it's mostly the same otherwise.


manhhung741:
Nice. Looking good. Since you already modify the code, why don’t we put the
generated sigs in a GeneratedUrlHelper module so that people can slap it on
any classes or modules to make the methods available?
ActionController::Base can include the module by default.

lib/sorbet-rails/routes_rbi_formatter.rb
```diff
@@ -4,62 +4,72 @@ class SorbetRails::RoutesRbiFormatter
...
-      # This is an autogenerated file for routes helper methods
-
-      # typed: strong
-      class ActionController::Base
-        extend T::Sig
-
-      #{@buffer.join("\n").indent(2)}
-      end
-    MESSAGE
+    @buffer = @parlour.rbi.split("\n")
```
manhhung741: It seems this should work with just @parlour.rbi
```suggestion
    @parlour.rbi
```

lib/sorbet-rails/routes_rbi_formatter.rb
```diff
@@ -4,62 +4,72 @@ class SorbetRails::RoutesRbiFormatter
 
   sig { void }
   def initialize
+    @parlour = T.let(Parlour::RbiGenerator.new, Parlour::RbiGenerator)
     @buffer = T.let([], T::Array[T.any(String, T::Array[String])])
```
manhhung741: You probably don't need @buffer any more. It's all to prepare data for the `result` method

lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -19,7 +19,7 @@ namespace :rails_rbi do
     root_dir = args[:root_dir] || Rails.root
     file_path = Rails.root.join("sorbet", "rails-rbi", "routes.rbi")
     FileUtils.mkdir_p(File.dirname(file_path))
-    File.write(file_path, inspector.format(SorbetRails::RoutesRbiFormatter.new))
+    File.write(file_path, inspector.format(SorbetRails::RoutesRbiFormatter.new).join("\n"))
```
manhhung741: Why is this needed? The whole content is generated in RoutesRbiFormatter.

connorshea: It was outputting a single line with a bunch of `\n` before, though I don't think that's true if we remove the buffer.

connorshea:
@manhhung741 Updated

lib/sorbet-rails/routes_rbi_formatter.rb
```diff
@@ -4,62 +4,71 @@ class SorbetRails::RoutesRbiFormatter
...
-  sig { params(routes: T.untyped).returns(NilClass) }
+  sig { params(routes: T.untyped).void }
   def header(routes)
+    @parlour.root.add_comments([
+      'This is an autogenerated file for Rails routes.',
+      'Please run rake rails_rbi:routes to regenerate.'
+    ])
+
+    @parlour.root.create_class('ActionController::Base') do |klass|
+      klass.create_include('GeneratedUrlHelper')
```
manhhung741:
```suggestion
      klass.create_include('GeneratedUrlHelpers')
```

lib/sorbet-rails/routes_rbi_formatter.rb
```diff
@@ -4,62 +4,71 @@ class SorbetRails::RoutesRbiFormatter
...
-        <<~MESSAGE
-          # Sigs for route #{r[:path]}
-          sig { params(args: T.untyped, kwargs: T.untyped).returns(String) }
-          def #{r[:name]}_path(*args, **kwargs); end
-          sig { params(args: T.untyped, kwargs: T.untyped).returns(String) }
-          def #{r[:name]}_url(*args, **kwargs); end
-        MESSAGE
-      else
-        nil
+    @parlour.root.create_module('GeneratedUrlHelper') do |klass|
```
manhhung741:
```suggestion
    @parlour.root.create_module('GeneratedUrlHelpers') do |klass|
```

manhhung741:
Ugh. There is still some string mismatch somewhere. \n vs no \n.
Can you run `./spec/bin/update_test_data.sh` ?

manhhung741:
I ran it but I don't have permissions to push to your PR branch

connorshea:
You should have permission, are you sure you’re not trying to force push?

connorshea:
BBC35E50-C160-4767-B452-83FF3C615108

manhhung741:
I'm pretty sure I didn't have permissions though -- like I couldn't commit the suggestions directly either

manhhung741:
I pushed the updated data

manhhung741:
Hmm, looks like there are still ruby version on the Gemfile of V6.0

manhhung741:
finally

connorshea:
👏

PR #126, opened 2019-08-15
Update files in spec/support/ apps

connorshea:
Mostly sorbet-typed-related changes. I was running the update Rake task anyway to test how long it took.

I wonder if we should just ignore sorbet-typed in `spec/support/` 🤔 The problem with that would be that we could miss sorbet-typed files that conflict with our definitions.


manhhung741:
I'm down to remove `sorbet-typed` and run `srb rbi sorbet-typed` in our `sorbet_spec.rb` It already run `srb rbi hidden-definitions`

PR #129, opened 2019-08-17
Redo 0.5.1 release

manhhung741:


manhhung741:
Turns out this is not needed

PR #130, opened 2019-08-18
Model relation runtime checkable

manhhung741:
We can make the following classes available to use at runtime:
- `Model::ActiveRecord_Relation` 
- `Model::ActiveRecord_AssociationRelation`
- `Model::ActiveRecord_Associations_CollectionProxy`

I also found out terrible truths in the process:
- The 3 above aren't subclass of each other -- you have to use the exact right one in each case
- When performing a query on CollectionProxy, it becomes AssociationRelation. Querying on an AssociationRelation & it's still assoc relation.

Because of that, I had to rework the generation logic related to relations to make all the operations on these classes work as expected at runtime.

In the process, I remove `ModelRelationShared` because it doesn't add any more value. Instead, we add query methods directly to 4 places:
- Model, as a class method
- Relation
- AssociationRelation
- CollectionProxy
This also help dealing with quirky issues with sorbet hidden-definitions -- the generated rbi sometimes overtake the method on generated rbis


lib/sorbet-rails/railtie.rb
```diff
@@ -15,5 +15,13 @@ class SorbetRails::Railtie < Rails::Railtie
       ActiveRecord::Base.extend SorbetRails::CustomFinderMethods
       ActiveRecord::Relation.include SorbetRails::CustomFinderMethods
     end
+
+    # in test & dev, the models are not pre-loaded, we need to load them manually
+    Rails.application.eager_load! unless Rails.env.production?
```
manhhung741: I wonder if there is a better way to do this :-? eg we can trigger it on any subclass of ActiveRecord::Base automatically

spec/generators/sorbet_test_cases.rb
```diff
@@ -11,37 +11,30 @@
 T.assert_type!(wizard.wand, T.nilable(Wand))
 T.assert_type!(T.must(wizard.wand).wizard, Wizard)
 
-T.assert_type!(wizard.spell_books, ActiveRecord::Associations::CollectionProxy)
-T.reveal_type(wizard.spell_books) # SpellBook::CollectionProxy
```
manhhung741: we don't need the reveal_type any more! yay!

PR #132, opened 2019-08-18
Include GeneratedUrlHelpers in ActionView::Helpers.

connorshea:
I did it this way because the URL Helpers aren't guaranteed to exist, so adding it to the helper generator would be kind of complicated.

I'm not sure if there's a better place to include these methods?

Essentially, I just want them available to me in `app/helpers/` files.


manhhung741:
> I did it this way because the URL Helpers aren't guaranteed to exist
Not only so, it's a dynamic module. You won't be able to statically put the module into the action view modules. I expect this to be the right way to make url helpers available in ActionView::Helpers

PR #133, opened 2019-08-19
Docs: known issue re: class methods on Relation

jaredbeck:
I'm documenting this known issue to prevent people from opening
new issues.


manhhung741:
Thanks!

PR #135, opened 2019-08-21
Check if method names are valid before adding

wpride:
When using `rails_rbi:models` I was running into an issue where `sorbet-rails` was generating invalid methods for accessing some of our `enum`s. These enums have names with spaces or apostrophes; for example, "Seller's Location". Usually the `enum` would create methods on the instance such as `Seller's Location?` and `Seller's Location!' as well as methods on the models to query for instances of that model matching that enum value. However, these methods do not exist and the resulting RBI files were invalid. 

This PR adds a simple regex to match valid Ruby method names. The criteria is that they begin with a letter, contain only letters and numbers, and can optionally end with a `!`, `?`, or `=`.

Please let me know if you think this is a good idea and I'll look into adding spec coverage. Thank you!


manhhung741:
Thanks for fixing the issue!

I think the named scope plugin should always give valid method name? The
method names comes from @class.methods itself. Maybe we have to escape the
name in some way?

You can get the list of methods for the models with funky enum & see what
methods actually generated for enums like that.

Alternatively, you can use the helper method method_exists? In model_utils
to test the method existence, instead of using regex.

For testing, you’d like to add a funky enum value to a model in
generators/rails-template then regenerate the rails app & the files.
Instructions are in Contributing.md.

manhhung741:
Would you like to follow up on this PR?

wpride:
@manhhung741 Updated specs with a "funky" enum. And we decided in Slack that the regular expression approach was best because these methods, despite having spaces, will still appear as methods on the class - right?

PR #137, opened 2019-08-22
Add SorbetRails.configure

connorshea:
Add the ability to configure what gem plugins are enabled.

Now you can do this in `config/initializers/sorbet_rails.rb`:

```ruby
SorbetRails.configure do |config|
  config.enabled_gem_plugins = [
    :kaminari
  ]
end
```

And it'll enable the plugin in your Rails app.

There are probably more elegant ways of doing this, but I wasn't able to get it working as just `SorbetRails.enabled_gem_plugins = []`. This is pretty much stolen from Kaminari's own `config` method :D https://github.com/kaminari/kaminari/blob/9e9ac3b7516b409a4aed1ce1bc6c5e90108c9511/kaminari-core/lib/kaminari/config.rb


lib/sorbet-rails/gem_plugins/gem_plugins.rb
```diff
@@ -0,0 +1,14 @@
+# typed: true
+require('sorbet-rails/config')
+
+module SorbetRails::GemPlugins
+  extend T::Sig
+
+  sig { void }
+  def self.register_gem_plugins
+    if SorbetRails.config.enabled_gem_plugins.include?(:kaminari)
```
connorshea: I should probably add a check here for whether the configured `enable_gem_plugins` includes only valid symbols, and error if there are any invalid symbols. Probably a better UX than doing nothing and never telling the user they've screwed up.

lib/sorbet-rails/config.rb
```diff
@@ -0,0 +1,28 @@
+# typed: true
```
manhhung741: strict?

lib/sorbet-rails/gem_plugins/gem_plugins.rb
manhhung741: we can do a switch:
```
SorbetRails.config.enabled_gem_plugins.each do |plugin_sym|
  case plugin_sym
  when :kaminari
     ...
  when :pg_search
    ...
  else
    InvalidPlugin.new
end
```

manhhung741:
I pushed some changes to consolidate the plugin configuration & also register plugins immediately after they are configured. Let me know what you think & feel free to make changes!

lib/sorbet-rails/model_plugins/plugins.rb
```diff
@@ -42,4 +30,43 @@ def set_plugins(plugins)
...
+
+  sig { params(plugin_name: Symbol).void }
+  def register_plugin_by_name(plugin_name)
+    register_plugin(get_plugin_by_name(plugin_name))
+  end
+
+  sig { params(plugin_name: Symbol).returns(T.class_of(Base)) }
+  def get_plugin_by_name(plugin_name)
+    case plugin_name
+    when :active_record_enum
```
connorshea: is the intention to allow disabling the core plugins?

manhhung741:
Yea, it’s intentional to be able to remove core plugins. I imagine in case
there is a conflict in a core plugin, people can remove it or subclass it
to deal with their edge case. This way people are not blocked when they
file issue/open PR for improvement.

connorshea:
Hmm, why is this failing with `uninitialized constant T` 🤔

PR #144, opened 2019-08-28
Make belongs_to relation non-nilable by default

wpride:
[Since Rails 5](https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html) the `belongs_to` relation has been required by default. To make this non-required you must add `optional: true` to the relation. 

This change updates the association methods to reflect this.

Curious what you all think about this change; if it looks good I will add tests.


manhhung741:
Sounds good. I verified that this is indeed the behavior in Rails 5. 
I'm curious though: what happen if you have a `belongs_to`, but the column is null? Is it just an ActiveRecord level check?

https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html

wpride:
@manhhung741 Yes, looks like you get an `ActiveRecord::NotNullViolation`:

Updated expected results

manhhung741:
Thanks!

PR #145, opened 2019-08-28
A couple of fixes for enums

deecewan:
There were a couple of things that didn't work _quite_ right with the enum types.

The first was the return type. It's not always an integer. Some places, including my company, use a string instead, to make the process of adding/removing enum values simpler and safer.

The second was the lack of support for prefix and suffix options. This would create types for methods that definitely didn't exist, and miss types for methods that definitely did exist.

Unfortunately, the only way I could think of to fix this was to cache every call to the `enum` method, and then can use it do statically determine if a prefix or suffix needs to be added. I followed the logic [here](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/enum.rb#L177) for determining the name.

Rails 4.2 doesn't support prefix or suffix. I've removed support for it, but unsure if that's allowed/recommended.


lib/sorbet-rails/model_plugins/active_record_enum.rb
```diff
@@ -13,20 +13,50 @@ def generate(root)
...
     model_class_rbi.create_include(enum_module_name)
 
+    enum_calls = ActiveRecordOverrides.instance.enum_calls[self.model_class_name]
+
     # TODO: add any method for signature verification?
     model_class.defined_enums.sort.each do |enum_name, enum_hash|
+      value_type = enum_hash.values.map(&:class).map(&:to_s).uniq
+
+      return_type = if value_type.length == 1
+        "T::Hash[T.any(String, Symbol), #{value_type.first}]"
```
manhhung741: Looks like we'll run into some issue with ruby 2.3 Integer vs FixNum hmm. IDK what's the best way to handle that.

spec/generators/rails-template.rb
```diff
@@ -114,28 +114,72 @@ def wood_type
...
+          chaser: 3,
+        }, _prefix: :quidditch
+
+        enum hair_color: {
+          brown: 0,
+          black: 1,
+          blonde: 2,
+        }, _suffix: :hair
+
+        enum eye_color: {
```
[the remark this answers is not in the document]
deecewan: sure! the point of this one was from some earlier testing, because the old system would have declared 2 `brown?` methods (one for hair, one for eyes) and I was trying to figure out a better way to get the methods. Now, tho, I'll add a prefix/suffix case

lib/sorbet-rails/model_plugins/active_record_enum.rb
deecewan: hmmm. it looks like the tests work as expected, but fail because 2.3 sees them as a `Fixnum` and the tests have an `Integer`.

I'm also not completely sure how to fix this...

manhhung741: We probably can manually change Fixnum -> Integer for Ruby 2.3

manhhung741:
I added a simple fix to tests with ruby 2.3. The PR now passes the tests!

deecewan:
very nice! i'll update the other issues today 😄

deecewan:
@manhhung741 this _should_ be complete now. let me know if there is anything else.

manhhung741:
Thanks. Looks really good! 
Ideally, we can add a few tests to sorbet_test_cases though:
- A test case that check enums with string element type
- A test case that check the method with prefix & suffix. This might be tricky since this only exists on 5.0 and above.

manhhung741:
Thank you! Great changes 🎉

PR #149, opened 2019-08-30
Add ActiveStorage attachment methods

### LINKED ISSUE #148: Generate methods for ActiveStorage attachments
**Describe the bug:**

SorbetRails doesn't currently generate methods for ActiveStorage attachments. For example, in my Rails app users have avatars and games have covers, but neither of these are typed.

**Steps to reproduce:**

```ruby
class User < ApplicationRecord
  has_one_attached :avatar
end
```

```ruby
class Game < ApplicationRecord
  has_one_attached :cover
end
```

In `hidden.rbi` we have some methods like this:

```ruby
module User::GeneratedAssociationMethods
  def avatar(); end

  def avatar=(attachable); end
end
```

**Expected behavior:**

It should return either [`ActiveStorage::Attached::One`](https://api.rubyonrails.org/classes/ActiveStorage/Attached/One.html) or [`ActiveStorage::Attached::Many`](https://api.rubyonrails.org/classes/ActiveStorage/Attached/Many.html) depending on the definition in the model.

This is what I'd expect to see in `game.rbi`:

```ruby
module Game::GeneratedAttributeMethods
  extend T::Sig
  
  sig { returns(T.nilable(ActiveStorage::Attached::One)) }
  def cover; end

  sig { params(attachable: T.untyped).returns(T.untyped) }
  def cover=(attachable); end

  # or, if we had something defined like `has_many_attached :covers`
  sig { returns(T.nilable(ActiveStorage::Attached::Many)) }
  def covers; end

  sig { params(attachable: T.untyped).returns(T.untyped) }
  def covers=(*attachables); end
end
```

I'm not entirely sure what the setters take or return, unfortunately.

**Versions:**
- Ruby: 2.6.2
- Rails: 6.0.0
- Sorbet: 
- Sorbet-Rails: 0.5.3 (master)

connorshea:
Also made some small formatting changes in a few places.

Fixes #148.

`attachment_reflections` wasn't added until 6.0, with https://github.com/rails/rails/pull/33018

So this doesn't support 5.2, and I'm not sure of a good way to get it to work. We can do something like `User.reflect_on_all_associations.filter { |x| x.name.to_s.end_with?('attachment') }`, but that's _really_ gross so I don't want to.

Are we fine just having this only support 6.0?

### DIFF
--- a/lib/sorbet-rails/config.rb
+++ b/lib/sorbet-rails/config.rb
@@ -49,8 +49,7 @@ module SorbetRails
         :enumerable_collections,
       ]
 
-      # TODO: Replace this with something sensible.
-      @enabled_model_plugins << :active_storage_methods if ['5.2', '6.0'].include?(ENV['RAILS_VERSION'])
+      @enabled_model_plugins << :active_storage_methods if defined?(T.unsafe(ActiveStorage))
     end
 
     sig { returns(T::Array[Symbol]) }
--- a/lib/sorbet-rails/config.rb
+++ b/lib/sorbet-rails/config.rb
@@ -48,6 +48,9 @@ module SorbetRails
         :custom_finder_methods,
         :enumerable_collections,
       ]
+
+      # TODO: Replace this with something sensible.
+      @enabled_model_plugins << :active_storage_methods if ['5.2', '6.0'].include?(ENV['RAILS_VERSION'])
     end
 
     sig { returns(T::Array[Symbol]) }
--- a/lib/sorbet-rails/model_plugins/active_storage_methods.rb
+++ b/lib/sorbet-rails/model_plugins/active_storage_methods.rb
@@ -8,7 +8,9 @@ class SorbetRails::ModelPlugins::ActiveStorageMethods < SorbetRails::ModelPlugin
 
   sig { implementation.params(root: Parlour::RbiGenerator::Namespace).void }
   def generate(root)
-    return unless @model_class.attachment_reflections.length > 0
+    # Check that ActiveStorage the attachment_reflections method exists
+    # It was added in 6.0, so it isn't available for 5.2.
+    return unless defined?(@model_class.attachment_reflections) && @model_class.attachment_reflections.length > 0
 
     assoc_module_name = self.model_module_name("GeneratedAssociationMethods")
     assoc_module_rbi = root.create_module(assoc_module_name)
--- a/lib/sorbet-rails/model_plugins/active_storage_methods.rb
+++ b/lib/sorbet-rails/model_plugins/active_storage_methods.rb
@@ -1,0 +1,58 @@
+# typed: true
+require ('sorbet-rails/model_plugins/base')
+class SorbetRails::ModelPlugins::ActiveStorageMethods < SorbetRails::ModelPlugins::Base
+  sig { params(model_class: T.class_of(ActiveRecord::Base), available_classes: T::Set[String]).void }
+  def initialize(model_class, available_classes)
+    super
+  end
+
+  sig { implementation.params(root: Parlour::RbiGenerator::Namespace).void }
+  def generate(root)
+    return unless @model_class.attachment_reflections.length > 0
+
+    assoc_module_name = self.model_module_name("GeneratedAssociationMethods")
+    assoc_module_rbi = root.create_module(assoc_module_name)
+
+    attachment_reflections = @model_class.attachment_reflections.transform_values { |attachment| attachment.class }
+
+    attachment_reflections.each do |assoc_name, attachment_type|
+      if attachment_type.to_s == 'ActiveStorage::Reflection::HasOneAttachedReflection'
+        create_has_one_methods(assoc_name, assoc_module_rbi)
+      elsif attachment_type.to_s == 'ActiveStorage::Reflection::HasManyAttachedReflection'
+        create_has_many_methods(assoc_name, assoc_module_rbi)
+      end
+    end
+  end
+
+  sig { params(assoc_name: String, mod: Parlour::RbiGenerator::Namespace).void }
+  def create_has_one_methods(assoc_name, mod)
+    mod.create_method(
+      assoc_name,
+      return_type: 'T.nilable(ActiveStorage::Attached::One)'
+    )
+
+    mod.create_method(
+      "#{assoc_name}=",
+      parameters: [
+        Parameter.new('attachable', type: 'T.untyped')
+      ],
+      return_type: 'T.untyped'
+    )
+  end
+
+  sig { params(assoc_name: String, mod: Parlour::RbiGenerator::Namespace).void }
+  def create_has_many_methods(assoc_name, mod)
+    mod.create_method(
+      assoc_name,
+      return_type: 'T.nilable(ActiveStorage::Attached::Many)'
+    )
+
+    mod.create_method(
+      "#{assoc_name}=",
+      parameters: [
+        Parameter.new('*attachables', type: 'T.untyped')
+      ],
+      return_type: 'T.untyped'
+    )
+  end
+end
--- a/spec/generators/rails-template.rb
+++ b/spec/generators/rails-template.rb
@@ -114,6 +114,12 @@ def create_models
     end
   RUBY
 
+  # A nasty hack to add has_one_attached and has_many_attached to the models/wizard.rb file.
+  attachments = nil
+  if ['5.2', '6.0'].include?(ENV["RAILS_VERSION"])
+    attachments = "has_one_attached :school_photo\n  has_many_attached :hats"
+  end
+
   if ENV["RAILS_VERSION"] == "4.2"
     file "app/models/wizard.rb", <<~RUBY
       class Wizard < ApplicationRecord
@@ -124,7 +130,7 @@ def create_models
           Hufflepuff: 1,
           Ravenclaw: 2,
           Slytherin: 3,
-	}
+        }
 
         enum professor: {
           "Severus Snape": 0,
@@ -193,6 +199,7 @@ def create_models
         has_many :spell_books
 
         scope :recent, -> { where('created_at > ?', 1.month.ago) }
+        #{attachments}
       end
     RUBY
   end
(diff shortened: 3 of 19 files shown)


spec/generators/rails-template.rb
```diff
@@ -129,6 +129,11 @@ class Wizard < ApplicationRecord
       has_many :spell_books
 
       scope :recent, -> { where('created_at > ?', 1.month.ago) }
+      
+      if ['5.2', '6.0'].include?(ENV["RAILS_VERSION"])
```
connorshea: I don't love this solution, but it's the least-worst I could come up with. I tried a few other ways to do this, and they were all pretty hacky and bad.

lib/sorbet-rails/model_plugins/active_storage_methods.rb
```diff
@@ -0,0 +1,60 @@
...
+  sig { implementation.params(root: Parlour::RbiGenerator::Namespace).void }
+  def generate(root)
+    # Check that ActiveStorage the attachment_reflections method exists
+    # It was added in 6.0, so it isn't available for 5.2.
+    return unless defined?(@model_class.attachment_reflections) && @model_class.attachment_reflections.length > 0
+
+    assoc_module_name = self.model_module_name("GeneratedAssociationMethods")
+    assoc_module_rbi = root.create_module(assoc_module_name)
+
+    attachment_reflections = @model_class.attachment_reflections.transform_values { |attachment| attachment.class }
```
connorshea: I guess I need to add `attachment_reflections` to the methods for ActiveRecord::Base in the `activerecord.rbi`?

connorshea:
Actually, here's a bit less disgusting option that works on Rails 5.2:
```ruby
ActiveStorage::Attachment.distinct.pluck(:record_type, :name)
#=> [["Game", "cover"], ["User", "avatar"]]
```

connorshea:
Oh also I have no idea how to distinguish between one and many associations with that ^

connorshea:
Tbh I think I’d rather just support this for Rails 6, Rails 5.2 will require too much bending over backwards

manhhung741:
I’m fine with supporting only 6.0 for now, as ActiveStorage really just
becomes first class API in 6.0. I can try to see if there is any
alternative to support 5.2 later.

lib/sorbet-rails/model_plugins/active_storage_methods.rb
manhhung741: Yeah, if there are sorbet typecheck errors without it

lib/sorbet-rails/config.rb
```diff
@@ -46,8 +46,10 @@ def initialize
         :active_record_assoc,
         :active_record_finder_methods,
         :custom_finder_methods,
-        :enumerable_collections,
+        :enumerable_collections
```
manhhung741: undo

lib/sorbet-rails/config.rb
```diff
@@ -46,8 +46,10 @@ def initialize
...
       ]
+
+      @enabled_model_plugins << :active_storage_methods if defined?(T.unsafe(ActiveStorage))
```
manhhung741: Hmm if we only support rails v6.0, we might need to change this condition.

connorshea: This doesn't really cause any problems since the plugin itself checks whether the given model has any `attachment_reflections`

connorshea:
Hmm, running the update spec script doesn't seem to change anything? Any ideas why the tests would be failing here?

manhhung741:
> Hmm, running the update spec script doesn't seem to change anything

Does the script run into any problem when regenerating data? It seems to remove `module Wizard::GeneratedAssociationMethods` entirely in your diff. I bet it run into a problem and skip the plugin. Try turning `allow_failure` to false in the `model_rbi_formatter`
```
  def run_plugins(plugins, generator, allow_failure: true)
```

connorshea:
Changing `allow_failure` doesn't seem to do anything either 🤔

connorshea:
The diff it expects is weird too:

```diff
@@ -105,28 +105,28 @@
module Wizard::GeneratedAssociationMethods
  extend T::Sig

-  sig { returns(::ActiveStorage::Attachment::ActiveRecord_Associations_CollectionProxy) }
+  sig { returns(ActiveRecord::Associations::CollectionProxy) }
  def hats_attachments; end

-  sig { params(value: T.any(T::Array[::ActiveStorage::Attachment], ::ActiveStorage::Attachment::ActiveRecord_Associations_CollectionProxy)).void }
+  sig { params(value: T.any(T::Array[T.untyped], ActiveRecord::Associations::CollectionProxy)).void }
  def hats_attachments=(value); end

-  sig { returns(::ActiveStorage::Blob::ActiveRecord_Associations_CollectionProxy) }
+  sig { returns(ActiveRecord::Associations::CollectionProxy) }
  def hats_blobs; end

-  sig { params(value: T.any(T::Array[::ActiveStorage::Blob], ::ActiveStorage::Blob::ActiveRecord_Associations_CollectionProxy)).void }
+  sig { params(value: T.any(T::Array[T.untyped], ActiveRecord::Associations::CollectionProxy)).void }
  def hats_blobs=(value); end

-  sig { returns(T.nilable(::ActiveStorage::Attachment)) }
+  sig { returns(T.nilable(T.untyped)) }
  def school_photo_attachment; end

-  sig { params(value: T.nilable(::ActiveStorage::Attachment)).void }
+  sig { params(value: T.nilable(T.untyped)).void }
  def school_photo_attachment=(value); end

-  sig { returns(T.nilable(::ActiveStorage::Blob)) }
+  sig { returns(T.nilable(T.untyped)) }
  def school_photo_blob; end

-  sig { params(value: T.nilable(::ActiveStorage::Blob)).void }
+  sig { params(value: T.nilable(T.untyped)).void }
  def school_photo_blob=(value); end

  sig { returns(::SpellBook::ActiveRecord_Associations_CollectionProxy) }
```

It generates `sig { returns(::ActiveStorage::Attachment::ActiveRecord_Associations_CollectionProxy) }` when I try to update the test data, but the tests seem to generate `sig { returns(ActiveRecord::Associations::CollectionProxy) }`?

manhhung741:
It must be having conflict with the association plugin hmm

connorshea:
Bizarrely, `spec/test_data/v5.2/expected_wizard_wo_spellbook.rbi` gets the correct sigs for `hats_attachment` but `spec/test_data/v5.2/expected_wizard.rbi` does not :?

spec/generators/rails-template.rb
connorshea: I've replaced this with a hack to get it to only add the lines for 5.2 and 6.0, but it's pretty gross. 🤷‍♂

connorshea:
ActiveStorage attachments and blobs are polymorphic, is that why it's acting so weird?

connorshea:
It looks like `relation_should_be_untyped?` is doing different things sometimes. Would `DISABLE_DATABASE_ENVIRONMENT_CHECK=1` matter?

connorshea:
Oh, actually when creating new instances of `SorbetRails::ModelRbiFormatter` the specs print `Note: Unable to create new instance of ActiveStorage::Attachment` locally but they're generated fine in CI. 🤔

connorshea:
@manhhung741 check d567df4ce7159bdebbe9aa0d83ac340084ce4b93 and see how much you hate it as a solution here :P

lib/sorbet-rails/model_plugins/active_storage_methods.rb
```diff
@@ -0,0 +1,60 @@
...
+  sig { params(model_class: T.class_of(ActiveRecord::Base), available_classes: T::Set[String]).void }
+  def initialize(model_class, available_classes)
+    super
+  end
+
+  sig { implementation.params(root: Parlour::RbiGenerator::Namespace).void }
+  def generate(root)
+    # Check that ActiveStorage the attachment_reflections method exists
+    # It was added in 6.0, so it isn't available for 5.2.
+    return unless defined?(@model_class.attachment_reflections) && @model_class.attachment_reflections.length > 0
```
manhhung741: Interesting that `defined?` work here. I'd think `respond_to?` or `method_defined?` makes more sense

spec/generators/rails-template.rb
```diff
@@ -128,7 +134,7 @@ class Wizard < ApplicationRecord
       has_one :wand
       has_many :spell_books
 
-      scope :recent, -> { where('created_at > ?', 1.month.ago) }
+      scope :recent, -> { where('created_at > ?', 1.month.ago) }#{attachments}
```
manhhung741: Is this change intentional? Looks like `#{attachments}` should be removed
```suggestion
      scope :recent, -> { where('created_at > ?', 1.month.ago) }
```

manhhung741:
I think I know what's wrong with the spec. This spec is what was failing. It actually uses the same expected file with a full run spec, which has the attachment & blob classes in "available_classes".
```
  it 'generates correct rbi file for Wizard' do
    formatter = SorbetRails::ModelRbiFormatter.new(Wizard, Set.new(['Wizard', 'Wand', 'SpellBook']))
    expect_match_file(
      formatter.generate_rbi,
      'expected_wizard.rbi',
    )
  end
```
I think your fix is correct.

spec/generators/rails-template.rb
connorshea: This is indeed correct, it’s the hack I used to generate the model with active storage methods in 5.2 and 6.0. You can’t put it on a new line or it causes blank lines in older versions

manhhung741: Can we put it in a new line? I'd prefer having an extra line than having attachments here on this line. People will be confused by it later for sure.

lib/sorbet-rails/model_plugins/active_storage_methods.rb
connorshea: Should I change it?

connorshea:
@manhhung741 I've rebased and addressed all your feedback.

manhhung741:
Thanks! I’ll take a look and merge later today

PR #151, opened 2019-09-02
Generate rbi for mailer classes

manhhung741:
For each instance mailing method, we detect the signature and generate the corresponding class method. 
- If a method has a sig, we'll generate a method with signature based on Sorbet's inflection data
- If a method doesn't have one, we'll generate a method with parameters all untyped
- The return type of a class method is `ActionMailer::MessageDelivery`, so that you can call "deliver_now", "deliver_later" on it.


jaredbeck:
Very cool, Harry! Can't wait to try it out.

Reading through the changes here was very educational for me, learning about how sorbet-rails is organized.

I have a feeling that the ability to reflect on method signatures will come in handy later.

PR #154, opened 2019-09-06
Allow including extra helpers as part of helpers.rbi

ghiculescu:
This is useful if you have extra global helpers apart from `Kernel` and `ActionView::Helpers`.


connorshea:
Nice! I've just been making the devise controller helpers accessible in helpers with a file called `custom.rbi` so far 🙈 

```ruby
# Make current_user available in GamesHelper and UsersHelper.
module GamesHelper
  include Devise::Controllers::Helpers
end

module UsersHelper
  include Devise::Controllers::Helpers
end
```

Your solution would be much nicer :)

manhhung741:
Nice. Thanks for the contribution! Currently there is a build failure with newest version of `sorbet`. I'm going to until it's fixed to merge the PR.

ghiculescu:
Ughhh, global config is super annoying. I'm pretty sure the config is leaking between tests, will try and fix properly tomorrow.

ghiculescu:
The remaining test also fails for me on master. I don't see how they are related to this PR. @manhhung741 	any tips?

connorshea:
@ghiculescu that’s what Harry was mentioning above, it’s a bug with Sorbet we need to fix. There’s nothing you need to do about it for now.

ghiculescu:
@connorshea 	ahh, I misread that because I also had tests failing. Thought it was just targeted at those. All good then, I'll sit tight until that's sorted!

manhhung741:
Haha, sorry for the confusion. I thought your PR passed the spec, but `sorbet_spec` was the only thing failing. Thanks for addressing the spec failures!

connorshea:
🎉 thanks!

PR #169, opened 2019-09-14
Add time zone config line to config/application.rb file and removing trailing blanks

jasnow:
* Add time zone config line to config/application.rb file for v5_0 version of Rails.
 * Remove trailing blanks from some files.

Welcome improvements and comments.


manhhung741:
Thanks for contributing!

Curious why is this only applied to Rails 5.0 and no other version?
@connorshea you have any opinion? You may have a better idea than me how best to add the config to Rails application.

Btw, I think you should reset the master branch with the current master of `sorbet-rails`. That way it doesn't show all the "merge" commits in the history

spec/generators/rails-template.rb
```diff
@@ -1,6 +1,25 @@
...
+     file "config/application.rb", <<~RUBY
+       require_relative 'boot'
+
+       require 'rails/all'
+
+       Bundler.require(*Rails.groups)
+
+       module TheApp
+         class Application < Rails::Application
+           config.active_record.time_zone_aware_types = [:datetime]
```
manhhung741: It may be simpler to have an initializer file that update the Raiils config.

jasnow:
Too much for this now so closing.

PR #171, opened 2019-09-14
Check enum_call nil before proceeding

manhhung741:
@wpride reported that sometimes `enum_call` could be nil for models. I'm unable to reproduce it but it's a good idea to be defensive around accessing nilable object.
cc @deecewan


wpride:
@manhhung741 Confirming this fixed the issue for me; thanks!

PR #178, opened 2019-09-15
Add helpers method

### LINKED ISSUE #156: Add support for helpers method in controllers
Rails 5.0+ has a method called `helpers` which is available in controllers. https://blog.bigbinary.com/2016/06/26/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers.html

So, for example if you had this `application_helper.rb`:

```ruby
module ApplicationHelper
  extend T::Sig

  sig { returns(Integer) }
  def forty_two
    return 42
  end
end
```

Then, in `users_controller.rb` you could have code like `helpers.forty_two`. `helpers` will include methods from any of the helper classes you've created.

```ruby
class UsersController < ApplicationController
  def index
    @number = helpers.forty_two
  end
end
```

It's not possible to create a type signature for this method statically, it has to be done dynamically.

e.g. like this:
```ruby
module ActionController::Helpers
  sig { returns(T.all(ApplicationHelper, GamesHelper, UsersHelper)) }
  def helpers; end
end
```

connorshea:
Resolves #156.

This intentionally does not include DeviseHelper because you can pass any kind of module into the `includes` config, and I don't think the Devise helpers are included in `helpers` anyway.

It checks for the existence of the `helpers` method because the method doesn't exist on 4.2.

### DIFF
--- a/lib/sorbet-rails/helper_rbi_formatter.rb
+++ b/lib/sorbet-rails/helper_rbi_formatter.rb
@@ -31,6 +31,16 @@ class SorbetRails::HelperRbiFormatter
       end
     end
 
+    # Adds the `helpers` method that provides access to all methods within the
+    # application's helpers.
+    # https://api.rubyonrails.org/classes/ActionController/Helpers/ClassMethods.html#method-i-helpers
+    @parlour.root.create_module('ActionController::Helpers') do |mod|
+      mod.create_method(
+        'helpers',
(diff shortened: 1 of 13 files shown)


manhhung741:
> It checks for the existence of the helpers method because the method doesn't exist on 4.2.

Soon 4.2 will not exist!

PR #179, opened 2019-09-15
Add 'new' method to models

### LINKED ISSUE #174: Define 'new' method on model classes
**Describe the bug:**
ActiveRecord overrides the `new` method for all Model classes, e.g. `User.new`.

It's defined like this in `activerecord.rbi`:
```ruby
module ActiveRecord::Inheritance::ClassMethods
  def new(attributes = nil, &block); end
end
```

And then `User`, or any other model, inherits that method.

**Expected behavior:**

I could have an `app/controllers/users_controller.rb` file like this:

```ruby
class UsersController < ApplicationController
  def create
    @user = User.new({ username: 'connor' }) # @user is T.untyped because `new` is overridden.
    respond_to do |format|
      if @user.save
        format.html { redirect_to @user, success: "#{@user.username} was successfully created." }
    end
  end
end
```

The method should be generated as part of Sorbet Rails so its return type can be dynamically set:

```ruby
class User
  sig { params(args: T.untyped, block: T.untyped).returns(User) }
  def self.new(*args, &block); end
end
```

Unfortunately it can't just be `self_type` because that doesn't return an instance of the caller :/

connorshea:
Resolves #174.

This takes the % of typed method calls (`sends_typed / sends`) from 70.18% to 71.06% for my Rails app 😎


manhhung741:
Can we create a new model_plugin for this? maybe `active_record_factory_methods`
While you're at it, can you also add sig for `create`, `create!` (other factory methods)?

connorshea:
The other factory methods are already typed via sorbet-typed, so I'm not sure if we should mess with them, necessarily. I guess we can override them regardless.

manhhung741:
I remember you said they are `T.untyped` in sorbet-typed? In
`sorbet-rails`, we can give them proper type.

connorshea:
They return untyped, but they do seem to have types for the parameters.

manhhung741:
Thank you! I appreciate the fix!

PR #180, opened 2019-09-16
`each` should return an array, not void

pje:
Currently the sigs we generate say that `each` returns `void`. This is incorrect—`each` always returns an array, just like `Enumerable#each`.

```
Wizard.first.spell_books.each(&:id).class
# => Array
```

This is because `each` actually gets delegated to `records`, which is just an array:

https://github.com/rails/rails/blob/v6.0.0/activerecord/lib/active_record/relation/delegation.rb#L85-L88

### DIFF
--- a/lib/sorbet-rails/model_plugins/enumerable_collections.rb
+++ b/lib/sorbet-rails/model_plugins/enumerable_collections.rb
@@ -36,6 +36,7 @@ class SorbetRails::ModelPlugins::EnumerableCollections < SorbetRails::ModelPlugi
       parameters: [
         Parameter.new("&block", type: "T.proc.params(e: #{self.model_class_name}).void")
       ],
+      return_type: "T::Array[#{self.model_class_name}]",
       implementation: true,
     )
     class_rbi.create_method(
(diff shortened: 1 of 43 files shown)


manhhung741:
Thanks @pje. This is correct, but I also think when people use `each`, they don't care about the result. `each` just return the same input array.

It's likely that what people want when they use the result of `each` is `map`. Can you explain the usecase you have when using the returning result of `each`?

frewsxcv:
> It's likely that what people want when they use the result of `each` is `map`. Can you explain the usecase you have when using the returning result of `each`?

The return type of `each` is an array, so the signature should reflect that. Can you expand on why the use-case matters?

pje:
I agree with @frewsxcv that correctness should be a primary goal in any code generation project—right now sorbet-rails is incorrectly generating sigs that cause type errors for perfectly valid ruby code, that's all I'm trying to fix.

I agree with you that _most of the time_, if someone's relying on the return value, it's better to create a new array using `map`. However, there are lots of reasons people might use `each` instead of `map`, but still rely on its return value. `each` doesn't create a new array, so it might be more performant. `each` can be inserted into a chain of transformations ([like `tap`'s example use case, but for a collection](https://ruby-doc.org/core-2.6.4/Object.html#method-i-tap)).

manhhung741:
Thanks for the explanation! I agree with what you are saying :-)

pje:
Thanks @manhhung741! Appreciate the merge. 🙇

manhhung741:
No, thank you for the fix :-)

PR #182, opened 2019-09-17
Set correct types for time, date and datetime attributes

clarkdave:
**Supports Rails 4.2+**

"Time like" attributes in ActiveRecord may be considered "time zone aware". This is the default in Rails 4.2 for `datetime` attributes, and from Rails 5.1 the default was extended to include `time` attributes too.

When an attribute is time zone aware, [its ActiveRecord type is decorated](https://github.com/rails/rails/blob/v6.0.0/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb#L76) and it will [begin casting its value](https://github.com/rails/rails/blob/v6.0.0/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb#L29) (which is usually a `DateTime` or `Time` instance) to an `ActiveSupport::TimeWithZone` instance when it's accessed or deserialized from the DB.

It's possible to disable this casting, [as described here](https://api.rubyonrails.org/classes/ActiveRecord/Timestamp.html). This PR takes these options into account by calling [a private class method](https://github.com/rails/rails/blob/v6.0.0/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb#L82) on the model with each "time like" attribute to determine if it's time zone aware or not. Unfortunately there is no public method that covers the full logic of this check, so the private method will have to do.

The end result is that all `date`, `datetime`, `timestamp` etc attributes should now be correctly typed as either `ActiveSupport::WithTimeZone` (if they're TZ-aware) or `Time` (if they're not).

As a convenience, TZ-aware attributes have their attribute writers typed as accepting a union type of `DateTime, Date, Time, ActiveSupport::TimeWithZone`. This is safe to do because Rails will always cast a TZ-aware attribute when it's assigned:

```rb
wand = Wand.first
wand.chosen_at_time.class # => ActiveSupport::TimeWithZone
wand.chosen_at_time = DateTime.now
wand.chosen_at_time.class # => ActiveSupport::TimeWithZone
```

---

In addition to the core functionality I regenerated all Rails apps (updating Sorbet in the process) and made a few changes to the `rails-template`:

- added a configuration option so that Rails 5.0 treats `time` columns as TZ-aware (this became the default behaviour in Rails 5.1 and makes the fixtures more consistent)
 - enforced not-null timestamps for `Wizard`, so the `created_at/updated_at` fields are consistently not null across all versions (allows for simpler `sorbet_test_cases`)
- added a `broken_at` column to `wands` table to support some additional assertions
- updated `Wand` model to ensure the new `broken_at` isn't TZ-aware for some assertions

### DIFF
--- a/lib/sorbet-rails/model_plugins/active_record_attribute.rb
+++ b/lib/sorbet-rails/model_plugins/active_record_attribute.rb
@@ -42,7 +42,7 @@ class SorbetRails::ModelPlugins::ActiveRecordAttribute < SorbetRails::ModelPlugi
         attribute_module_rbi.create_method(
           "#{column_name}=",
           parameters: [
-            Parameter.new("value", type: column_type.to_s)
+            Parameter.new("value", type: value_type_for_attr_writer(column_type))
           ],
           return_type: nil,
         )
@@ -61,7 +61,11 @@ class SorbetRails::ModelPlugins::ActiveRecordAttribute < SorbetRails::ModelPlugi
       ActiveRecord::Base.connection.lookup_cast_type_from_column(column_def) :
       column_def.cast_type
 
-    strict_type = active_record_type_to_sorbet_type(cast_type)
+    strict_type =
+      active_record_type_to_sorbet_type(
+        cast_type,
+        time_zone_aware: time_zone_aware_column?(column_def, cast_type),
+      )
 
     if column_def.respond_to?(:array?) && column_def.array?
       strict_type = "T::Array[#{strict_type}]"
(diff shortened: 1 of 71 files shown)


lib/sorbet-rails/model_plugins/active_record_attribute.rb
```diff
@@ -107,4 +112,41 @@ def active_record_type_to_sorbet_type(klass)
...
+  def time_zone_aware_column?(column_def, cast_type)
+    # this private class method returns true if the attribute should be "time
+    # zone aware"; it takes into account various global and model-specific
+    # configuration options as described here:
+    # https://api.rubyonrails.org/classes/ActiveRecord/Timestamp.html
+    #
+    # although it's private, it's better this than trying to reimplement the "is
+    # this attribute tz aware?" logic ourselves
+    @model_class.send(
+      :create_time_zone_conversion_attribute?,
```
clarkdave: Although private, this method has been in place for **10 years**, so as private methods go, it seems quite stable!

https://github.com/rails/rails/blame/55cc16f502399b69051e4ff5432569bbea45e9c0/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb

manhhung741:
Thank you! The changeset looks very good!

PR #185, opened 2019-09-19
Call active_record_type recursively for Postgres Arrays

wpride:
Currently all of our array columns have type `T::Array[T.untyped]`. I guess this is because we can't introspect on the type of the array's members without knowing the specific SQL implementation used.

This change handles the Postgres Array case specifically, calling the `active_record_type_to_sorbet_type` on the `subtype` when found. 

This works for our models, but maybe this is looking into the cookie jar too much for the library?


manhhung741:
Can you fix spec & add specs to illustrate the change?

manhhung741:
Closing because this PR has been outdated for a long time. @wpride If you still want to make the change, please send another PR based on the newest code

PR #187, opened 2019-09-21
Add sigs for ActiveRecord::Core module

donaldong:
Add signatures for some methods in [ActiveRecord::Core](https://api.rubyonrails.org/classes/ActiveRecord/Core.html).

Since not all methods can be statically typed easily, this module is still in `todo.rbi`.


manhhung741:
Hey Donald, I think this is more fitting adding to sorbet-typed gem.

https://github.com/sorbet/sorbet-typed

Also this activerecord.rbi that you edit is only for type-checking
sorbet-rails code itself. We don’t bundle an activerecord.rbi with
sorbet-rails.

PR #189, opened 2019-09-26
Fix Typo link to file config.rb in the README.md

udn:
Hi there, I found typo on link to config.rb file, the typo is :
- before : https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/lib/sorbet-rails/config.rb 
- should be : https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/config.rb


manhhung741:
Thank you!

PR #190, opened 2019-09-27
Allow symbols to be assignable for AR attributes

mgriisser:
The params for attribute assignment to string DB columns appears to be overly restrictive. 

In principle, I think, you can assign almost any `Object` to these columns, e.g. `model.attr = Object.new` gives something like `attr:  "#<Object:0x00007f94b66f3d10>"`. But I'm not sure a change that general makes sense.

Probably should allow symbols though given their ubiquity and similarity to strings.


manhhung741:
Thanks for the PR! It makes a lot of sense to be able to assign Symbol to a string attribute.
 
Currently we use the type as both attribute type & assignable type, but Rails do a lot of "conversion" at the assignable type. Maybe we can look up the logic in rails for type conversion and try to mimic that for the general case.

manhhung741:
Can you fix the spec failures? You can run `./spec/bin/update_test_data.sh` to re-record the tests

mgriisser:
Done. Thanks for the heads up on the test regen script! I thought there must be something like that but hadn't had time to look into it/ask about it.

manhhung741:
Thank you!

Issue #152, opened 2019-09-03
Select has incorrect return type when passed a block


riley-klingler:
**Describe the bug:**
[ApplicationRecord.select](https://apidock.com/rails/v5.2.3/ActiveRecord/QueryMethods/select) has two different modes: one that takes a block and returns an Array, and one that takes any number of fields and returns an `ActiveRecord::Relation`. The `sig` that `sorbet-rails` generates claims it can only return a `Relation`.

**Steps to reproduce:**
```
class User < ApplicationRecord
end
```
Generated `rbi` file:
```
  sig { params(args: T.untyped, block: T.nilable(T.proc.void)).returns(User::ActiveRecord_Relation) }
  def select(*args, &block); end
```
**Expected behavior:**
To be honest, I'm not quite sure. Making the return type of `select` be `T.any(User::ActiveRecord_Relation, T::Array[User])` is not great, because callers will always have to check which type is being returned, even though it can always be inferred from the argument type. But it might be the best we can do, since ruby doesn't support overloading.

**Versions:**

- Ruby: 2.6.2
- Rails: Rails 5.2.3
- Sorbet:  0.4.4651
- Sorbet-Rails: 0.5.0

manhhung741:
Hi Riley, like you said, `sorbet` doesn't support having 2 sig for one method. The `.select` method that returns an array is a select performed on the materialized element of an active record.

You can do this for typechecking instead: `User.all.to_a.select` <- this will take a block and returns an Array

riley-klingler:
Do you want to remove the block as a parameter then? Right now, the `sig` produced implies that you can pass in a block and return an `ActiveRecord_Relation`, which will not happen. It also implies you can pass both `args` and a block, which will [throw a runtime error](https://github.com/rails/rails/blob/bdfd47053108f3e7757f35a6d3eee511bf6e0cc5/activerecord/lib/active_record/relation/query_methods.rb#L256).

manhhung741:
Ah darn. Would you like to send a PR to remove &block and add an item to
“Tips and Tricks” about how to use `select` correctly?

`*args, &block` is the standard param for a bunch of delegate and/or
flexible methods. There might be more methods we can remove &block.

pje:
> Making the return type of select be T.any(User::ActiveRecord_Relation, T::Array[User]) is not great, because callers will always have to check which type is being returned, even though it can always be inferred from the argument type.

[...]

PR #191, opened 2019-09-30
Remove block parameter from querying methods

riley-klingler:
Some of these methods (e.g. `left_joins`) do not take a block at all. Others (e.g. `select`) can take a block, but `sorbet` is not expressive enough to [properly typecheck](https://github.com/chanzuckerberg/sorbet-rails/issues/152) calls where a block is passed in.

I checked all of these functions and think it's correct that either of those two conditions are true for all of them, but checking was fairly repetitive so it would be great if someone could check my work.

### DIFF
--- a/lib/sorbet-rails/model_plugins/active_record_querying.rb
+++ b/lib/sorbet-rails/model_plugins/active_record_querying.rb
@@ -24,7 +24,7 @@ class SorbetRails::ModelPlugins::ActiveRecordQuerying < SorbetRails::ModelPlugin
     # rails/activerecord/lib/active_record/querying.rb
     model_query_relation_methods = [
       :select, :reselect, :order, :reorder, :group, :limit, :offset, :joins, :left_joins, :left_outer_joins,
-      :where, :rewhere, :preload, :extract_associated, :eager_load, :includes, :from, :lock, :readonly, :extending, :or,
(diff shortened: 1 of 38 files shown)


manhhung741:
Thank you Riley! I'll review the PR asap!

lib/sorbet-rails/model_plugins/active_record_querying.rb
```diff
@@ -33,7 +33,6 @@ def generate(root)
         method_name.to_s,
         parameters: [
           Parameter.new("*args", type: "T.untyped"),
-          Parameter.new("&block", type: "T.nilable(T.proc.void)"),
```
manhhung741: I think all querying methods are defined here:

https://github.com/rails/rails/blob/9a7f15c905c9ca7a26b56b6f2a045cf56805a99e/activerecord/lib/active_record/relation/query_methods.rb

I did a grep of block and find that other than `select`, `extending` is the only other method accepting a block
https://github.com/rails/rails/blob/9a7f15c905c9ca7a26b56b6f2a045cf56805a99e/activerecord/lib/active_record/relation/query_methods.rb#L918

Can you separate it from the rest?

riley-klingler: Thanks for catching that!

manhhung741: Looks good now! thanks for fixing!

PR #194, opened 2019-10-02
Add warning about using with `sorbet-runtime`

ghiculescu:
Maybe this is overzealous, but we got really burned by it. We were using `sorbet-rails` in development only, but `sorbet-runtime` in all environments. This meant we were able to access `X::ActiveRecord_Relation` in development, and thus someone used that in a method signature. When we deployed this code it failed in production because [this railtie](https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/railtie.rb#L30) didn't run, so the constant was still private.

Adding a warning so it doesn't happen to other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ gem 'sorbet-rails'
 ❯ bundle install
 ```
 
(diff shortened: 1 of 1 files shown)


manhhung741:
There are things I add at runtime like `find_n`, `first_n`, etc. as well so likely you always want it to be in production mode. What do you think about warning people to always use it in production?

ghiculescu:
That seems simpler. I can reword tomorrow to just say that.

manhhung741:
Thank you!

manhhung741:
And sorry for the issue you ran into

ghiculescu:
All good, it was our own fault for not testing more thoroughly. I simplified the message.

manhhung741:
Thank you!

PR #206, opened 2019-10-15
Fix `belongs_to` required check.

frewsxcv:
A `belongs_to` relationship is only required in the following cases:

> - `required` is specified on the association AND it's `true`
> 
> OR
> 
> - `optional` is specified on the association AND it's `false`
> 
> OR
> 
> - `model.belongs_to_required_by_default` is `true`

This logic [matches ActiveRecord' logic](https://github.com/rails/rails/blob/fd6648d5979df6e6daf9fa7eb1d01cd87922ae32/activerecord/lib/active_record/associations/builder/belongs_to.rb#L109-L125) for determining whether to add a presence validation for the `belongs_to` association.

Prior to this pull request, sorbet-rails would ignore the `belongs_to_required_by_default` option, and would incorrectly treat `optional: nil` as `optional: false`, and wouldn't factor in `required`.


lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -35,12 +35,7 @@ def generate(root)
   def populate_single_assoc_getter_setter(assoc_module_rbi, assoc_name, reflection)
     # TODO allow people to specify the possible values of polymorphic associations
     assoc_class = assoc_should_be_untyped?(reflection) ? "T.untyped" : "::#{reflection.klass.name}"
-    assoc_type = "T.nilable(#{assoc_class})"
-    if reflection.belongs_to?
-      column_def = @columns_hash[reflection.foreign_key.to_s]
```
frewsxcv: I removed this since it didn't seem to have any effect

manhhung741:
Is it possible to change a spec model to have a `belong_to` but it's not required?

manhhung741:
The change looks good to me btw. I'm considering releasing a new version for recent fixes after merging this PR :-)

frewsxcv:
> Is it possible to change a spec model to have a `belong_to` but it's not required?

Hmm, not sure I understand. Can you reword the question? What's a spec model?

manhhung741:
I mean can we add a spec to test the code change. You can change code in a
model in the spec, eg Wizard model.

frewsxcv:
I started writing a test but ran out of time. I may not be able to finish it for a while, FYI

manhhung741:
Let me try to patch the spec

manhhung741:
Test fixed!

PR #209, opened 2019-10-17
Fix bug with usage of `present?`

frewsxcv:
Follow-up to https://github.com/chanzuckerberg/sorbet-rails/pull/206

```
> reflection.options
=> {:inverse_of=>:payment_intents, :optional=>false}
```

```
> reflection.options[:optional].present?
=> false
```

```
> reflection.options.key?(:optional)
=> true
```


manhhung741:
Uh, so like there are people who use `optional: false` as opposed to `required: true`?

frewsxcv:
> Uh, so like there are people who use `optional: false` as opposed to `required: true`?

Yep!

manhhung741:
Thanks for the fix! Let me know if you want me to release a .1 version to include it.

Issue #211, opened 2019-10-22
Running tests raises a NameError for sigs receiving or returning an ActiveRecord_AssociationRelation


franciscoj:
**Describe the bug:**

When you declare a method which receives or returns an `ActiveRecord_AssociationRelation` of some sort, tests fail with:

```
     NameError:
       private constant #<Class:0x00007fa75c4e8d28>::ActiveRecord_AssociationRelation referenced
```

**Steps to reproduce:**

create a method which returns an association relation on a model

```ruby
class Post < ApplicationRecord
  belongs_to :user

  scope :published ->() { where(published: true) }
end

class User < ApplicationRecord
  extend T::Sig

  has_many :posts

  sig { returns(Post::ActiveRecord_AssociationRelation) }
  def published_posts
    posts.published
  end
end

RSpec.describe User do
  describe '#published_posts' do
    it 'returns only the published posts' do
      user = User.create!
      published = user.posts.create(published: true)
      user.posts.create(published: false)

      expect(user.published_posts).to eq([published])
    end
  end
end
```

This would raise a `NameError` claming that `Post::ActiveRecord_AssociationRelation` is a private class.

Note as well that in my case the models are on a rails engine.

**Expected behavior:**

The test passes

**Versions:**

- Ruby: 2.4
- Rails: 5.2.3
- Sorbet: Sorbet typechecker 0.4.4891 git 4c9a4534439483b46e869c7212007ad103923203 built on 2019-10-17 22:56:24 GMT debug_symbols=true
- Sorbet-Rails: 0.5.8

manhhung741:
Can I ask if sorbet-rails is included in your test and production
environment? Those classes are Rails private class. We have logic that is
loaded through initializer to make them publicly accessible. Usually this
error you see means they aren't publicly available.

franciscoj:
You're totally right.

I think I've managed to fix it by moving `gem 'sorbet-rails'` on the `Gemfile` from the `group :development` block to the general group.

Thanks a lot!

International:
I have sorbet-rails in the general group, but I receive the same issue with private constant being referenced.

franciscoj:
I can confirm that moving it to the general group fixed the issue for me. Besides the code is now running on CI and on production with some `MyModel::ActiveRecord_Association` typed methods.

@International what I did to see what was happening on my case was to:

```
bundle open sorbet-rails
```

Go to the railties file and add a `byebug` on the inherited hook.

[...]

PR #214, opened 2019-10-25
Docs: signatures for methods returning relations

jaredbeck:
I couldn't find any docs on this, so I'm attempting to write some. I'm not at all sure this is correct, but will at least start a conversation. Thanks!


manhhung741:
Jared, we do support using Bacon::ActiveRecord_Relation. I do agree with
need better documentation for it

jaredbeck:
> we do support using Bacon::ActiveRecord_Relation ..

Oh, great! Should I update this PR to make that the recommendation?

manhhung741:
Sure, I think we need it. Hopefully we can resolve the issue like with #211 

Thanks for helping with the documentation. Here is some thought I've had:
- We need a whole section calling out things `sorbet-rails` provide at runtime (as opposed to things provided via the static generator). It includes following functionality:
+ Relation class: Making the relations available at runtime (they are private, the gem make them public)
+ Model: find_n, first_n, last_n are the methods we added. pluck_to_tstruct as well.
+ Controller: adding fetch_typed and require_typed

- And make sure people put `sorbet-rails` in a common group in their Gemfile (not development or test only)

jaredbeck:
> Here is some thought I've had

Nice! Thanks Harry. I've added your thoughts, and re-organized a little.

README.md
```diff
@@ -19,16 +19,14 @@ This gem adds a few Rake tasks to generate Ruby Interface (RBI) files for dynami
 
 ```
 # -- Gemfile --
-
+# In the default group (same as sorbet-runtime) because of "Features Provided at Runtime" below
```
manhhung741: I've heard some people don't use `sorbet-runtime` in the production environment, so I think this part may actually be misleading. Maybe we should say "the group including production environment"

README.md
```diff
@@ -148,6 +146,22 @@ Since mailing action methods is based on instance methods defined in a mailer cl
 - If not, all the params in the mailing action method will be T.untyped.
 - For return type though, the mailing action method will return `ActionMailer::MessageDelivery` instead of the return type of the instance method.
 
+## Features Provided at Runtime
```
manhhung741: I think we should change the current "Type-checking Rails code" to "Static RBI Generation", and move the documentation of run-time features there to this section.

manhhung741: What's needed for this section to work properly is running `rake rails_rbi:custom`, which will copy the rbi for the provided methods.

jaredbeck:
Updated per comments, please review.

> What's needed for this section to work properly is running rake rails_rbi:custom, which will copy the rbi for the provided methods.

I don't understand this comment. Is it something you can add in a future PR?

README.md
```diff
@@ -148,6 +147,23 @@ Since mailing action methods is based on instance methods defined in a mailer cl
 - If not, all the params in the mailing action method will be T.untyped.
 - For return type though, the mailing action method will return `ActionMailer::MessageDelivery` instead of the return type of the instance method.
 
+### Features Provided at Runtime
```
manhhung741:
```suggestion
## Runtime Features
```

manhhung741:
Btw, please remove `[ci skip]` from your commit message. Github checks the PR passes the Travis CI check before it can be merged. Our repo is set up to requires PR passes CI run to merge. Even though this change doesn't make any code change, it's easier to have consistent rule on this.

manhhung741:
> I don't understand this comment. Is it something you can add in a future PR?
Sure, I'll add it later. I'll attempt to explain it still: simply put we need to copy some RBI so that the sorbet knows the signature of the runtime methods added. It's through running the rake task
```
rake rails_rbi:custom
```

jaredbeck:
Thanks Harry, changes made.

PR #223, opened 2019-11-11
Integer/Float attributes accept many different types

ghiculescu:
Currently we are quite strict about what types can be passed to an attribute of type `Integer`/`Float`, but in reality Rails is quite accommodating. This PR allows more types to be passed in.

Fixes https://github.com/sorbet/sorbet-typed/pull/163#issuecomment-552647540


lib/sorbet-rails/model_plugins/active_record_attribute.rb
```diff
@@ -131,15 +131,23 @@ def time_zone_aware_column?(column_def, cast_type)
 
   sig { params(column_type: Object).returns(String) }
   def value_type_for_attr_writer(column_type)
```
manhhung741: We probably should pass a `column_def` to this method and have a switcher based on `active_record_type` like `active_record_type_to_sorbet_type`

manhhung741:
Created an issue for code quality improvement.

PR #226, opened 2019-11-15
Handle NotImplementedError raises

sbeckeriv:
Dearest Reviewer,

I hope all is well this November.  It as been a long month for me and I look forward to the amazing things sorbet will help me with. I have interface classes that raise NotImplementedError and the model process is dying on it. I have found locally this to be an acceptable change. I hope that you agree. 

Changes
Add NotImplementedError to the model processing exception handling for interface like classes.

To reproduce add a method that raises NotImplementedError. 

Becker


manhhung741:
Thank you for the change! I'm merging it now.

manhhung741:
FWIW, `NotImplementedError` is usually used for features not implemented in the system-level. It looks like some gem/library used this for their own stuff. I don't think it's necessarily wrong or harmful, just want to call it out. I think for our team, we define a different error equivalent to `NotImplementedError`.

PR #236, opened 2019-12-11
Handle belongs_to association that is null: false in the DB

manhhung741:
Talked to @palasha about something we do to belongs_to in our codebase: we rely in DB-level configuration to decide whether it's required or optional. Our code doesn't specify a required or optional config for belongs_to association, but if the column is not nullable & there is a foreign-key constraint between the two tables, the association should not be `nilable`. We also think DB constraint is stronger than application-level constraint.

I make this change so that we generate non-nilable association when at the DB level, the column backing the `belongs_to` connection is not-nilable, and warn user when the config at the DB level is less strict than the application level.


lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -56,13 +56,25 @@ def populate_single_assoc_getter_setter(assoc_module_rbi, assoc_name, reflection
...
+      end
+
+    column_def = @columns_hash[reflection.foreign_key.to_s]
+    db_required_config = column_def && !column_def.null
+
+    if rails_required_config && !db_required_config
+      puts "Warning: belongs_to association #{reflection.name} is required at the application
+        level but not at the DB level. It's recommended to enforce it by adding null & foreign key
+        constraint at the DB level.".squish!
     end
```
ghiculescu: how do you feel about also adding a warning for the inverse case?

```ruby
    elsif !rails_required_config && db_required_config
      puts "Warning: belongs_to association #{reflection.name} is required at the DB
        level but not at the application level. Add a constraint at the app level (using `optional: false`)
        as a validation hint to Rails.".squish!
    end
```

manhhung741: I can, but when you do both, it also means doing extra unnecessary validation. What do you think?
 
I consider this direction of the warning because the DB level is the most certain level to put any kind of restriction (application level is not sufficient). Having both probably means doing extra work.

manhhung741: Looking at the implementation of `belongs_to` and `validates_presence_of`, I think it's worthy to have both: the validation is simple (`.blank?`) and probably cost a little to do. Lack of it means people may not be truly aware of the required-ness of the association. 
I'll add a warning whenever the two configurations don't match.

ghiculescu: We've been removing `validates_presence_of` in favour of `belongs_to optional: false` because of the sorbet-rails benefits it gives. Thank you for adding this!

manhhung741:
Thanks for reporting. I’ll look into it today!

manhhung741:
PR is out: https://github.com/chanzuckerberg/sorbet-rails/pull/242

PR #242, opened 2019-12-13
Fix belongs to and required returns nil

manhhung741:
@ghiculescu found a bug related to `belongs_to_and_required` return `nil` instead of T::Boolean. I fix it, but I also feel that this method, or methods that return T::Boolean should probably return T.nilable(T::Boolean), because it's very common in ruby/rails to return `nil` for boolean-y functions. Mmmm


ghiculescu:
Thank you!

PR #244, opened 2019-12-19
12 13 fix belongs to and required returns nil

Sa1Beer777:
12 13 fix belongs to and required returns nil


manhhung741:
@Sa1Beer777  is this PR created by mistake? This branch is already merged to master.

PR #249, opened 2019-12-26
Add ActiveRecord::Relation#empty?

mikker:
`any?` is the opposite of ... https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/active_record/relation.rb#L265-L268


manhhung741:
Thank you :-) and happy new year!

PR #256, opened 2020-01-17
Add a few more static attribute alternatives

ghiculescu:
These came up as common errors while updating our internal Sorbet docs.


manhhung741:
Thank you, @ghiculescu !

PR #296, opened 2020-03-14
Adds XXX::RelationType alias

masylum:
Many times we just want to check that something receives a collection of
models. Internally Rails uses 3 different classes to represent
an enumerable interface to ActiveRecord.

This commit adds the generation of the following alias:

```rb
class User
  ActiveRecord_Enumerable = T.type_alias do
    T.any(
      User::ActiveRecord_Relation,
      User::ActiveRecord_AssociationRelation,
      User::ActiveRecord_Associations_CollectionProxy
    )
  end
end

### DIFF
--- a/lib/sorbet-rails/railtie.rb
+++ b/lib/sorbet-rails/railtie.rb
@@ -34,6 +34,16 @@ class SorbetRails::Railtie < Rails::Railtie
             child.send(:public_constant, :ActiveRecord_Relation)
             child.send(:public_constant, :ActiveRecord_AssociationRelation)
             child.send(:public_constant, :ActiveRecord_Associations_CollectionProxy)
+
+            relation_type = T.type_alias do
+              T.any(
+                child.const_get(:ActiveRecord_Relation),
+                child.const_get(:ActiveRecord_AssociationRelation),
+                child.const_get(:ActiveRecord_Associations_CollectionProxy)
+              )
+            end
+            child.const_set(:RelationType, relation_type)
+            child.send(:public_constant, :RelationType)
           end
         end
       end
--- a/lib/sorbet-rails/model_rbi_formatter.rb
+++ b/lib/sorbet-rails/model_rbi_formatter.rb
@@ -109,6 +109,10 @@ class SorbetRails::ModelRbiFormatter
     )
     model_rbi.create_extend("T::Sig")
     model_rbi.create_extend("T::Generic")
+    model_rbi.create_type_alias(
+      self.model_relation_type_class_name,
+      type: self.model_relation_type_alias
+    )
   end
 
   sig {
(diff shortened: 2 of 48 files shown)

### CODE REVIEW

**Comment by manhhung741** [comment]
I think we also need to generate ruby implementation so that the aliases
are usable at runtime. Either we generate a ruby file with the alias (and
tell user to load the file), or we should hook into Railtie and generate it
dynamically. For the latter, you can check out the code where we make the
Relation classes public. It might be a good place to also create the alias.

Naming-wise, this alias represents more than just Enumerable. I wonder if
we should name it like User::RelationType

**Comment by manhhung741** [comment]
Thanks for creating the PR btw :-) I think the aliases will be super useful!

**Comment by masylum** [comment]
Hi @manhhung741! I wasn't expecting you to see this yet, I was just playing around hehe and getting to see how the project works.

> I think we also need to generate ruby implementation so that the aliases
are usable at runtime. Either we generate a ruby file with the alias (and
tell user to load the file), or we should hook into Railtie and generate it
dynamically. For the latter, you can check out the code where we make the
Relation classes public. It might be a good place to also create the alias.

Cool! Will keep investigating over this weekend :)

> Naming-wise, this alias represents more than just Enumerable. I wonder if
we should name it like User::RelationType

Sounds good to me 👍

**Comment by manhhung741** [comment]
Thank you :-) Feel free to let me know if you have any question about the
code!

File: lib/sorbet-rails/railtie.rb [hunk 1]
```diff
@@ -34,6 +34,7 @@ def inherited(child)
             child.send(:public_constant, :ActiveRecord_Relation)
             child.send(:public_constant, :ActiveRecord_AssociationRelation)
             child.send(:public_constant, :ActiveRecord_Associations_CollectionProxy)
+            child.send(:public_constant, :RelationType) if child.const_defined?(:RelationType)
```
manhhung741: This constant doesn't exist and we'll have to define it for the child.
Something like 
```
# define the type alias dynamically
# if we are unable to use ActiveRecord_Relation, we would have to get it manually
# active_record_relation = const_get(:ActiveRecord_Relation)
relation_type = T.type_alias { 
  ActiveRecord_Relation, 
  ActiveRecord_AssociationRelation, 
  ActiveRecord_Associations_CollectionProxy
}
# create a constant mapping to the type alias.
child.const_set(:RelationType, relation_type)
```

masylum: ok, I have to admit that this file was a bit beyond my understanding... 😅

manhhung741:
@masylum FYI I added a spec and helped you fix the missing rbi

btw we're in "shelter in place" mode in the bay area now too :(

manhhung741:
Hmm I should add some documentation

PR #303, opened 2020-04-02
better handling of habtm associations

### LINKED ISSUE #299: HABTM relationships not handled correctly
first i wanted to say thanks for this gem and specifically this typechecking between AR and the DB via reflection. it's awesome and it helped us catch a bunch of potential issues recently. 

**Steps to reproduce:**

given a model with a HABTM relationship where the column is `NOT NULL` in the db, we see this output:
```
Note: belongs_to association company is specified as not-null at the DB level but **optional** at the application level. Add a constraint at the app level (using `optional: false`) as a validation hint to Rails.
```

**Expected behavior:**

as far as i can tell there's no way to do this for HABTM, so the messaging should be changed but i'm not sure to what? happy to open a pr to help with this. 

**Versions:**
- Ruby: ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin19]
- Rails: Rails 5.2.4.1
- Sorbet: Sorbet typechecker 0.5.5469 git 42d26158c0a6ae45eb5a0e8eb71c53616190ae2e debug_symbols=true clean=1
- Sorbet-Rails: 0.6.1

modosc:
fixes #299


modosc:
oops, looking at the build failure now.

modosc:
ugh - i've tried rebuilding everything locally and can't get that single spec to pass, any ideas?

manhhung741:
@modosc Hmm. Looking. It seems something with Subject model not being loaded when that test run. Because of that, the generation logic uses "T.untyped" instead of Subject.

manhhung741:
You can also run the spec locally. If it passes locally but fails in Travis, I'll look more into why.
```
bundle exec rspec ./spec/model_rbi_formatter_spec.rb:17
```

lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -73,9 +73,16 @@ def populate_single_assoc_getter_setter(assoc_module_rbi, assoc_name, reflection
         level but **nullable** at the DB level.\n Add a constraint at the DB level
         (using `null: false` and foreign key constraint) to ensure it is enforced.".squish!
     elsif !rails_required_config && db_required_config
-      puts "Note: belongs_to association #{reflection.name} is specified as not-null at the
-        DB level but **optional** at the application level.\n Add a constraint at the app level
-        (using `optional: false`) as a validation hint to Rails.".squish!
+      if habtm?
+        puts "Note: belongs_to association #{reflection.name} is specified as not-null at the
```
manhhung741, line 77: These messages try to be helpful in terms of suggesting good configuration. I think we likely don't want to remove `not-null` constraint at the DB level. Perhaps we can skip logging for this case.

lib/sorbet-rails/model_utils.rb
```diff
@@ -5,27 +5,42 @@ module SorbetRails::ModelUtils
...
+  def habtm?
+    # checking the class name seems to be the cleanest way to figure this out, see:
+    # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/associations/builder/has_and_belongs_to_many.rb#L54
+    T.must(model_class.name).start_with?('HABTM_')
+  end
+
   sig { returns(String) }
   def model_class_name
-    "#{model_class.name}"
+    if habtm? && model_class.respond_to?(:left_model)
```
manhhung741, line 22: Perhaps we can make it applied to any nested models, because I think this logic is applicable to more things than just HABTM classes. Eg, we can differentiate:
- Model full name (User::HABTM_xxx)
- Model name (HABTM_xxx)

In most cases, we would want to use model full name. And it can be retrieved using "to_s":
```
User::HABTM_Roles.to_s
=> "User::HABTM_Roles"

irb(main):017:0> User::HABTM_Roles.name
=> "HABTM_Roles"
```

lib/sorbet-rails/tasks/rails_rbi.rake
```diff
@@ -133,12 +133,16 @@ namespace :rails_rbi do
   def generate_rbis_for_models(model_classes, available_classes)
     available_class_names = Set.new(available_classes.map { |c| c.name })
     formatted = model_classes.map do |model_class|
+      # NOTE: this logic is duplicated in SorbetRails::ModelUtils#model_class_name
```
manhhung741, line 136: again, here we can use "model_class.to_s" to get the class name including the namespace.

lib/sorbet-rails/model_utils.rb
modosc: hah, `to_s` is way simpler. will update everywhere.

lib/sorbet-rails/model_plugins/active_record_assoc.rb
modosc: this warning would have been useful for my employer's codebase but i don't feel that strongly about it. 

fwiw i'm also working on updating the rails docs about `habtm`

lib/sorbet-rails/tasks/rails_rbi.rake
modosc: done

lib/sorbet-rails/model_plugins/active_record_assoc.rb
modosc: or alternately i can remove the suggestion about removing the not null constraint and leave the other suggestion (which is what we did locally)?

modosc:
ok - fixed the build/spec issues, i think the message is the last item to sort out?

lib/sorbet-rails/model_plugins/active_record_assoc.rb
manhhung741: Yeah, let's remove the suggestion about "removing not null constraint"

lib/sorbet-rails/model_utils.rb
```diff
@@ -5,27 +5,36 @@ module SorbetRails::ModelUtils
...
   abstract!
 
-  sig { abstract.returns(T.class_of(ActiveRecord::Base)) }
+  # if we're a HABTM class then model_class is an anonymous class (see the rails link below) and
+  # i'm not sure how to explain that to sorbet other than T.class_of(Class).
+  sig { abstract.returns(T.any(T.class_of(ActiveRecord::Base), T.class_of(Class))) }
   def model_class; end
 
+  sig { returns(T::Boolean) }
+  def habtm?
```
manhhung741, line 14: nit: "habtm_class?"

modosc: done

modosc:
ok we should be good now.

manhhung741:
@modosc FYI I've released a new version including this change.

PR #304, opened 2020-04-07
Collection IDs

ghiculescu:
Rails generates methods of the form `singular_collection_name_ids` for has_many relationships. This PR adds those methods into the sorbet-rails generated RBIs.


ghiculescu:
@manhhung741 

https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html

image

there's also a setter which I could probably add too

ghiculescu:
@manhhung741 this one's green now :)

connorshea:
Just out of curiosity, wouldn't this also be able to return an array of strings if the ID column were a UUID?

PR #305, opened 2020-04-14
Use percent-string instead of single quotes

Gasparila:
For `T::Enum` enums generated from rails enums, use `%q{}` instead of
`''` so that enums with single-quotes won't break sorbet-rails

Also added a test case which before generated a bunch of errors and now works

### DIFF
--- a/spec/generators/rails-template.rb
+++ b/spec/generators/rails-template.rb
@@ -149,6 +149,7 @@ def create_models
         "Pomona Sprout": 2,
         "Filius Flitwick": 3,
         "Hagrid": 4,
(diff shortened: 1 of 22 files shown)


spec/generators/rails-template.rb
```diff
@@ -149,6 +149,7 @@ class Professor; end
         "Pomona Sprout": 2,
         "Filius Flitwick": 3,
         "Hagrid": 4,
+        "Alastor 'Mad-Eye' Moody": 5,
```
manhhung741, line 152: 👁️‍🗨️

manhhung741:
Thank you!!!

PR #306, opened 2020-04-14
Use `presence` validations to remove T.nilable on attributes and associations

jeffcarbs:
This PR updates the model RBI generator for column attributes, belongs_to, and has_one to check for `presence` validations to help determine if we can avoid using `T.nilable`.

For example let's say we have a `Student` model with the following schema:
```
#  id          :bigint(8)    not null, primary key
#  first_name  :string(255)  not null
class Student
  validates :first_name, presence: true
end
```

In this case, `first_name` has a non-null DB constraint so we know the type of `first_name` is `String`. Let's pretend it there was no constraint:
```
#  id          :bigint(8)    not null, primary key
#  first_name  :string(255)
class Student
  validates :first_name, presence: true
end
```

In the model RBI that sorbet-rails generates the `first_name` would be `T.nilable(String)` which means anywhere we use first_name we need to check for nil to make the static checker happy. However, there's an `presence: true` validation on first_name which means that we can expect `first_name` to in fact not be nilable (ignoring the validation being added after some data had already been entered or columns were updated via `update_columns` to skip the validation). This PR makes use of the validation to tighten the signature to `String`.

NOTE: This only works for _unconditional_ validations (i.e. ones without `:if` or `:unless`). Here's an example that would still be treated as nilable:
```
#  id          :bigint(8)    not null, primary key
#  first_name  :string(255)
class Student
  validates :first_name, presence: true, unless: :anonymous?
end
```

Because we can't statically be sure that the validation will be called we can't confidently set the type to `String`.

We also do this for:
- belongs_to associations (we check for validations on both the association name OR the foreign key)
- has_one associations (we check for validations on the association name)


lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -65,8 +68,7 @@ def populate_single_assoc_getter_setter(assoc_module_rbi, assoc_name, reflection
         !!reflection.active_record.belongs_to_required_by_default
       end
 
-    column_def = @columns_hash[reflection.foreign_key.to_s]
-    db_required_config = column_def.present? && !column_def.null
+    rails_required_config ||= [column_def&.name, reflection.name].compact.any? { |n| attribute_has_unconditional_presence_validation?(n) }
```
manhhung741, line 71: can you add a comment here why we can use "reflection.name"? is it so that it checks the validation logic for both object (wizard) and column attribute (wizard_id)?

lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -137,4 +144,13 @@ def polymorphic_assoc?(reflection)
       polymorphic_assoc?(reflection.source_reflection) :
       reflection.polymorphic?
   end
+
+  sig { params(attribute: T.any(String, Symbol)).returns(T::Boolean) }
+  def attribute_has_unconditional_presence_validation?(attribute)
```
manhhung741, line 149: should we move this logic to the base plugin to share the code?

sorbet/rbi/todo.rbi
```diff
@@ -76,3 +76,14 @@ module T::Private::Methods::MethodHooks; end
 module T::Private::Methods::Signature; end
 module T::Private::Methods::SingletonMethodHooks; end
 module T::Props::Plugin::ClassMethods; end
+
+class ActiveModel::Validations::PresenceValidator
```
manhhung741, line 80: you can add this to the rbi in dependent_gem_rbis?
https://github.com/chanzuckerberg/sorbet-rails/blob/master/lib/sorbet-rails/dependent_gem_rbis/activerecord.rbi
(that file or you can create activemodel.rbi)

manhhung741: This file may be overridden by sorbet when we regenerate rbi for new sorbet versions

spec/test_data/v5.1/expected_headmaster.rbi
```diff
@@ -0,0 +1,681 @@
...
+  sig { returns(Integer) }
+  def id; end
+
+  sig { params(value: T.any(Numeric, ActiveSupport::Duration)).void }
+  def id=(value); end
+
+  sig { returns(T::Boolean) }
+  def id?; end
+
+  sig { returns(T.nilable(Integer)) }
```
manhhung741, line 22: I thought this attribute has validation and should use non-nilable value as return type?

spec/test_data/v5.1/expected_headmaster.rbi
```diff
@@ -0,0 +1,681 @@
...
+
+  sig { returns(T::Boolean) }
+  def wizard_id?; end
+end
+
+module Headmaster::GeneratedAssociationMethods
+  extend T::Sig
+
+  sig { returns(::School) }
+  def school; end
```
manhhung741, line 45: eg it can detect that "school" is non-nilable here.

spec/test_data/v5.0/expected_wizard.rbi
```diff
@@ -167,10 +167,10 @@ module Wizard::GeneratedAttributeMethods
   sig { returns(T::Boolean) }
   def id?; end
 
-  sig { returns(T.nilable(String)) }
+  sig { returns(String) }
```
manhhung741, line 170: nice

spec/test_data/v5.1/expected_headmaster.rbi
jeffcarbs: Oh, this is interesting. In the example `school` has a presence validation and that `belongs_to` logic is happening in the `ActiveRecordAssoc`. If the association were not nilable then theoretically the FK should also be not nilable but we don't consider the association in the `ActiveRecordAttribute` plugin.

This would be the same behavior as the existing logic around `belongs_to_and_required` since that's only handled in `ActiveRecordAssoc` plugin, not the attribute one. Given that, would you be OK not handling this in the current PR and I could maybe follow-up with something that would update `ActiveRecordAttribute` to inspect associations (i.e. look at validations but also optional/required)?

jeffcarbs:
Made the updates you suggested and asked if we could defer one of them (see https://github.com/chanzuckerberg/sorbet-rails/pull/306#discussion_r409262196).

PR #307, opened 2020-04-16
Generate RBIs for ActiveRecord::Base and ActiveRecord::Relation

jeffcarbs:
`active_record_base.rbi` and `active_record_relation.rbi` were previously maintained by hand but they have a lot of the same methods that ought to be kept in sync (e.g. the finder methods like `find`, `first`, etc). This PR adds a `ActiveRecordRbiFormatter` that generates these RBIs.

For the most part I tried to keep the diff as small as possible. I first generated the `expected_*` files and then separated out the base and relation changes to their own commits. git also thinks that `lib/bundled_rbi/active_record_base.rbi` was renamed to `spec/test_data/v5.0/expected_active_record_base.rbi` which should make the diff pretty easy to see.

There are some more fixes/updates that should be made to these RBIs but I think I'll save that for a new PR to keep them easier to review.


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  sig {
+    params(
+      class_rbi: Parlour::RbiGenerator::Namespace,
+      type: String,
+      class_method: T::Boolean,
+    ).void
+  }
+  def create_finder_methods(class_rbi, type, class_method)
```
manhhung741, line 99: Can you make type & class_method keyword arguments? (esp "class_method"). It'll help with reading & comprehending the callsites

lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -0,0 +1,175 @@
...
+        "flatten",
+        parameters: [ Parameter.new("level", type: "T.nilable(Integer)") ],
+        return_type: "T::Array[Elem]",
+      )
+      # this is an escape hatch when there are conflicts in signatures of Enumerable & ActiveRecord
+      class_rbi.create_method(
+        "to_a",
+        return_type: "T::Array[Elem]",
+      )
+      # TODO use type_parameters(:U) when parlour supports it
```
manhhung741, line 64: parlour supports it now if you would like to update the code^^

lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -0,0 +1,175 @@
...
+          def map(&blk); end
+        RUBY
+      )
+
+      class_rbi.create_method(
+        "exists?",
+        parameters: [ Parameter.new("conditions", type: "T.untyped", default: "nil") ],
+        return_type: "T::Boolean",
+      )
+      class_rbi.create_method('any?', return_type: "T::Boolean")
```
manhhung741, line 82: nice!

lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -0,0 +1,175 @@
...
+      class_rbi.create_method(
+        item_method,
+        return_type: item_method.ends_with?('!') ? type : "T.nilable(#{type})",
+        class_method: class_method,
+      )
+    end
+
+    build_methods = %w(create create! new)
+    build_methods.each do |build_method|
+      # This should be defined on both but trying to keep the diff small in this commit
```
manhhung741, line 137: Thanks for thinking about the reviewers here! Feel free to send out another PR to make changes

spec/rake_rails_rbi_active_record_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    expect_files(
+      base_dir: generated_dir_path,
+      files: [
+        "active_record_base.rbi",
+        "active_record_relation.rbi",
+      ],
+    )
+  end
+end
```
manhhung741, line 21: nit: add a newline

jeffcarbs:
Addressed the feedback. Will follow up with a PR to add some of the missing method definitions.

PR #308, opened 2020-04-17
Update ActiveRecord Base and Relation RBI generators

jeffcarbs:
### Background

The reason you can call things like `Wizard.count` is because ActiveRecord::Base delegates a ton of methods to `:all` (see https://github.com/rails/rails/blob/master/activerecord/lib/active_record/querying.rb#L5-L21).

For RBI-sake, you can categorize them into three categories:
1. Returns primitives (T::Boolean, Integer, etc)
2. Returns the model itself or an array of models (e.g. Wizard or T::Array[Wizard])
3. Returns an ActiveRecord::Relation specific to the model (e.g. Wizard::ActiveRecord_Relation)

For category 1, you can actually define them in a common module that gets included in `ActiveRecord::Base` and `ActiveRecord::Relation` verbatim.

For category 2, you can define them on both `ActiveRecord::Base` and `ActiveRecord::Relation`. They can't share an RBI since `ActiveRecord::Base` needs to use class methods and return `T.attached_class` and `ActiveRecord::Relation` needs to use instance methods and return `Elem`.

For category 3, you must define them within the model RBI itself because those need to use the relation classes specific to the models like `Wizard::ActiveRecord_Relation`.

### This PR

Category 1 and 2 were mostly handled but some common ones (at least in our codebase) were missing so I added them. For now the missing generic methods were added to both the `Base` and `Relation` to keep the diff smaller but we could could DRY it up in a future PR by moving those to a separate module that `Base` would `extend` and `Relation` would `include`.

One other thing to point out is that the `activerecord.rbi` that is generated by sorbet (in sorbet/rbi/gems) defines a `ActiveRecord::Associations::CollectionProxy` and `ActiveRecord::AssociationRelation` which define methods like `create` which causes sorbet use that one (which doesn't have sigs) instead of the one inherited from `ActiveRecord::AssociationRelation`. For the methods that got overwritten I added them explicitly to the definitions of CollectionProxy and AssociationRelation.


lib/sorbet-rails/model_plugins/enumerable_collections.rb
```diff
@@ -1,71 +0,0 @@
...
-    if include_methods
-      class_rbi.create_method(
-        "each",
-        parameters: [
-          Parameter.new("&block", type: "T.proc.params(e: #{self.model_class_name}).void")
-        ],
-        return_type: "T::Array[#{self.model_class_name}]",
-        implementation: true,
-      )
-      class_rbi.create_method(
```
manhhung741, line 47: I think this method (flatten) is missing from the new generation code. Can you add it?

jeffcarbs: I don't think `flatten` is a method on ActiveRecord::Relation or related classes:
```ruby
> User.all.flatten
NoMethodError: undefined method `flatten' for #<User::ActiveRecord_Relation:0x00007f8c7eb904d0>
```

manhhung741: But if it's part of an array, it works 🤷 
```
> [User.all].flatten
```

jeffcarbs:
Closing since these changes were chunked up into small PRs.

PR #309, opened 2020-04-20
Improve array typing

Gasparila:
Add better typing for array types to include the array subtype and also
fix the nullability. Unfortunately there is no easy way to add tests as
the current framework (understandably) uses SQLite for the DB engine
which does not support array types. I tested this with some of our data
models, however, and saw this worked properly:

sample `git diff` for table with `assignee_names` as `:string, array:
true`:
```
-  sig { returns(T::Array[T.nilable(T.untyped)]) }
+  sig { returns(T.nilable(T::Array[String])) }
   def assignee_names; end

-  sig { params(value: T::Array[T.nilable(T.untyped)]).void }
+  sig { params(value: T.nilable(T::Array[T.any(String, Symbol)])).void }
   def assignee_names=(value); end
```

Note that I only tested this on PostgreSQL so I'm unsure if this will fully work
on other ActiveRecord adapters. As a result, I added a defensive `if cast_type.try(:subtype)`
so that, at worst, the new behavior will be the same if other adapters implement
arrays differently (e.g. don't use `subtype`)

### DIFF
--- a/lib/sorbet-rails/model_plugins/active_record_attribute.rb
+++ b/lib/sorbet-rails/model_plugins/active_record_attribute.rb
@@ -12,8 +12,9 @@ class SorbetRails::ModelPlugins::ActiveRecordAttribute < SorbetRails::ModelPlugi
     sig { returns(String) }
     def to_s
       type = base_type.to_s
-      type = "T.nilable(#{type})" if nilable
+      # A nullable array column should be T.nilable(T::Array[column_type]) not T::Array[T.nilable(column_type)]
       type = "T::Array[#{type}]" if array_type
+      type = "T.nilable(#{type})" if nilable
       type
     end
   end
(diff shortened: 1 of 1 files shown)


lib/sorbet-rails/model_plugins/active_record_attribute.rb
```diff
@@ -12,8 +12,9 @@ class ColumnType < T::Struct
     sig { returns(String) }
     def to_s
       type = base_type.to_s
-      type = "T.nilable(#{type})" if nilable
+      # A nullable array column should be T.nilable(T::Array[column_type]) not T::Array[T.nilable(column_type)]
       type = "T::Array[#{type}]" if array_type
+      type = "T.nilable(#{type})" if nilable
```
manhhung741, line 17: thanks for catching this bug

Gasparila: Eh..... I wrote it 😄

manhhung741: 😅 Well I'm glad you come back and fix it

PR #310, opened 2020-04-22
Add quotations to auto-generated rails_rbi command

wpride:
`zsh` users cannot copy and paste the command in `Please rerun bundle exec rake rails_rbi:models[Shipment] to regenerate.` without escaping the brackets. Adding comments around the command makes this runnable with `zsh` and `bash` (and looks nicer).


manhhung741:
Thanks, @wpride! It seems you only change the generation code in "helper_rbi_formatter". Do you want to also change it for other formatter classes? and I think you need to rerun the generation code as well.

lib/sorbet-rails/helper_rbi_formatter.rb
```diff
@@ -18,7 +18,7 @@ def generate_rbi
 
     @parlour.root.add_comments([
       'This is an autogenerated file for Rails helpers.',
-      'Please rerun bundle exec rake rails_rbi:helpers to regenerate.'
+      'Please rerun \"bundle exec rake rails_rbi:helpers\" to regenerate.'
```
manhhung741, line 21: Because we use` ' ' `here, you don't need to escape the `"`. Currently your change generate this:
```
       +# Please rerun \"bundle exec rake rails_rbi:helpers\" to regenerate.
```
But I think we only want?
```
       +# Please rerun "bundle exec rake rails_rbi:helpers" to regenerate.
```

PR #311, opened 2020-04-23
Add more sorbet test cases

jeffcarbs:
- Add more assertions to the `sorbet_test_cases.rb`. Most have been commented out since sorbet-rails doesn't currently support them.
- Minor updates to `ActiveRecordRbiFormatter` to support some of the new assertions.

This is the first of a few PRs that I'll be spinning out of #308 to make it easier/possible to review.

### DIFF
--- a/spec/generators/sorbet_test_cases.rb
+++ b/spec/generators/sorbet_test_cases.rb
@@ -62,6 +62,26 @@ T.assert_type!(Wizard.find_by(name: 'Harry Potter'), T.nilable(Wizard))
 T.assert_type!(Wizard.find_by!(name: 'Harry Potter'), Wizard)
 T.assert_type!(Wizard.find_by_id(wizard.id), T.nilable(Wizard))
 T.assert_type!(Wizard.find_by_id!(wizard.id), Wizard)
+T.assert_type!(Wizard.find_or_initialize_by(name: 'Harry Potter') { |w| T.assert_type!(w, Wizard) }, Wizard)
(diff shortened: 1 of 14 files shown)


spec/generators/sorbet_test_cases.rb
```diff
@@ -114,48 +186,38 @@
...
-
-# Enumerable on activerecord relation
-T.assert_type!(Wizard.all.to_a, T::Array[Wizard])
-Wizard.all.each do |w|
-  T.assert_type!(w, Wizard)
-end
-Wizard.all.map do |w|
-  T.assert_type!(w, Wizard)
-end
-Wizard.all.to_a.map do |w|
```
jeffcarbs, line 132: This assertion wasn't super useful since we already had an assertion testing that `.to_a` would return a `T::Array[Wizard]`.

manhhung741: The first one is useful because sometimes the "map" or the "each" function is overridden by Sorbet autogen rbi. I have it here to make sure that "each" and "map" work correctly.

manhhung741: Some of them, like the check `Wizard.all.map` are there to also make sure that sorbet autogen rbi doesn't make it stop working for subclasses.

However, I see you also have these checks above. We can remove them here.

jeffcarbs: Yea I kept those but moved them into the sections for relation, association relation, and collection association. I just removed this one test for them since we were already ensuring that to_a returned an array of things, so we shouldn’t need to test that to_a.each has the correct behavior unless we’re skeptical of sorbet’s sigs for builtin ruby methods.

manhhung741:
Yeah I agree.

PR #312, opened 2020-04-24
Generate sigs for more build/create ActiveRecord methods

jeffcarbs:
Follow-up to #311. Generates RBI for `build`, `first_or_create`, `first_or_create!`, `first_or_initialize`.

NOTE: The first 3 commits here are from that branch, they'll go away once that is merged and I rebase.

### DIFF
--- a/lib/sorbet-rails/active_record_rbi_formatter.rb
+++ b/lib/sorbet-rails/active_record_rbi_formatter.rb
@@ -127,13 +127,16 @@ class SorbetRails::ActiveRecordRbiFormatter
       )
     end
 
-    build_methods = %w(create create! new)
+    build_methods = %w(create create! new build first_or_create first_or_create! first_or_initialize)
     build_methods.each do |build_method|
+      # `build` method doesn't exist on the model, only on the relations
+      next if build_method == 'build' && class_method
+
       # This needs to match the generated method signature in activerecord.rbi and
       # in Rails 5.0 and 5.1 the param is a splat.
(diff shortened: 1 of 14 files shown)


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -127,13 +127,16 @@ def create_elem_specific_query_methods(class_rbi, type:, class_method:)
...
     build_methods.each do |build_method|
+      # `build` method doesn't exist on the model, only on the relations
+      next if build_method == 'build' && class_method
+
       # This needs to match the generated method signature in activerecord.rbi and
       # in Rails 5.0 and 5.1 the param is a splat.
       if Rails.version =~ /^5\.(0|1)/ && %w(new build create create!).include?(build_method)
         param = Parameter.new("*args", type: "T.untyped")
-      elsif
+      else
```
manhhung741, line 139: Oh was this a bug that I didn't notice? I already merged the other PR so lets try to merge this one quickly

manhhung741:
There are some conflicts because I squash the changes from #311 into 1 commit in master. You can rebase and drop the first 3 commits to resolve the conflicts.

lib/sorbet-rails/active_record_rbi_formatter.rb
jeffcarbs: It definitely wasn't intended but it would still work. I think ruby evaluated the next line as the `elsif` statement but only if we didn't enter the first branch so it actually behaved the same.

jeffcarbs:
@manhhung741 - rebased and ready to review 👍

PR #313, opened 2020-04-26
Generate sigs for build/create methods on AssociationRelation and CollectionProxy

jeffcarbs:
Follow up to #312. The first two commits in this PR are from that one.

`ActiveRecord::AssociationRelation` and `ActiveRecord::Associations::CollectionProxy` both inherit from `ActiveRecord::Relation` so adding sigs for methods on `Relation` should trickle to the others. However, when sorbet autogenerates RBI via reflection is defines an `activerecord.rbi` which defines a few methods on `AssociationRelation` and `CollectionProxy` that take precedence over the inherited ones meaning that we lose sig information.

To get around this, we need to explicitly add some methods to our RBI which is what this PR is doing.

### DIFF
--- a/lib/sorbet-rails/active_record_rbi_formatter.rb
+++ b/lib/sorbet-rails/active_record_rbi_formatter.rb
@@ -81,6 +81,34 @@ class SorbetRails::ActiveRecordRbiFormatter
       class_rbi.create_method('one?', return_type: "T::Boolean")
     end
 
+    parlour.root.create_class("ActiveRecord::AssociationRelation", superclass: "ActiveRecord::Relation") do |class_rbi|
+      class_rbi.create_constant(
+        "Elem",
+        value: "type_member(fixed: T.untyped)",
+      )
+
+      # Ideally we shouldn't need to define these since this class inherits from
+      # ActiveRecord::Relation but the activerecord.rbi that sorbet generates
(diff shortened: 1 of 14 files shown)


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -81,6 +81,62 @@ def generate_active_record_relation_rbi
...
+      # defines some methods which sorbet finds instead of the methods inherited
+      # by ActiveRecord::Relation. Some of these methods have different arity or
+      # parameters than the ones defined by `create_elem_specific_query_methods` so
+      # we need to match the signatures in that conflicting rbi.
+      build_methods = %w(new build create create!)
+      build_methods.each do |build_method|
+        class_rbi.create_method(
+          build_method,
+          parameters: [
+            Parameter.new("*args", type: "T.untyped"),
```
connorshea, line 101: I think this is wrong? It was changed to have `attributes` just like the proxy methods: https://github.com/rails/rails/commit/f1c1c43022c24e27a02d698fd29e7955c10d2f03

The sorbet-rails tests fail now if you run the tests and it installs Rails 6.0.3. (I'm not sure if 6.0.2/6.0.1 are effected?)

PR #314, opened 2020-04-26
Move methods into base AssociationRelation and CollectionProxy RBIs

jeffcarbs:
Follow up to #312 and #313. The first two commits in this PR are from the first one one, the next two are from the second.

This PR moves the `finder` methods from directly in the model RBIs to being generated in a base class (usually `ActiveRecord::Relation` although some needed to be explicitly defined in `CollectionProxy`).

Going by commit:
- Remove usage of `typed-override` in specs (pulled out into #315)
- Added missing assertions for the "numbered finders" (e.g. first, second, third)
- Remove the generation of most `finder_methods` from the model RBIs
- Regenerate the model RBIs


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -135,6 +135,29 @@ def generate_active_record_relation_rbi
...
+        parameters: [Parameter.new("*args", type: "T.untyped")],
+        return_type: "Elem",
+      )
+
+      if Rails.version =~ /^5\.0/
+        item_methods = %w(first second third third_to_last second_to_last last)
+        item_methods.each do |item_method|
+          class_rbi.create_method(
+            item_method,
+            parameters: Rails.version =~ /^5\.0/ ? [Parameter.new("*args", type: "T.untyped")] : [Parameter.new("limit", type: "T.untyped", default: "nil")],
```
manhhung741, line 150: This seems to be the same check as the outer "if"?

spec/test_data/v5.1/expected_active_record_relation.rbi
```diff
@@ -162,4 +162,10 @@ class ActiveRecord::Associations::CollectionProxy < ActiveRecord::Relation
 
   sig { params(attributes: T.untyped, block: T.nilable(T.proc.params(object: Elem).void)).returns(Elem) }
   def create!(attributes = nil, &block); end
+
+  sig { params(args: T.untyped).returns(Elem) }
```
manhhung741, line 166: Interesting that these methods are only needed for `ActiveRecord::Associations::CollectionProxy`. Is it working for `ActiveRecord::Relation` and `ActiveRecord::AssociationRelation` because they use the signatures from L34 to L59?

lib/sorbet-rails/active_record_rbi_formatter.rb
jeffcarbs: You're right, nice catch 👍

jeffcarbs:
@manhhung741 - Rebased and fixed that issue re: the redundant if statement.

To your other thought/question:
> I don't even think it's possible to reason about when a method can be inherited from ActiveRecord::Relation and when we need to regenerate it.

If sorbet did not auto-generate an RBI for activerecord, we wouldn't need to do this. Here's a snippet from `spec/support/v5.0/sorbet/rbi/gems/activerecord.rbi`:
Screen Shot 2020-04-26 at 7 20 13 PM

What's happening is that sorbet is seeing:
- ActiveRecord::Relation define methods like `build`, `create`, etc
- ActiveRecord::Associations::CollectionProxy inherit from ActiveRecord::Relation so it now will have those methods with the sigs we defined in Relation
- ActiveRecord::Associations::CollectionProxy defines some of those methods (e.g. highlighted in red in the screenshot) and so those take precedence but unfortunately they don't have sigs.

The `sorbet_test_cases.rb` now tests everything we're writing sigs for in those base classes (Base, Relation, etc) and that they're correctly typed on the model-specific classes so we basically just have to rely on that to know when we need to explicitly define methods vs falling back to the inherited ones.

spec/test_data/v5.1/expected_active_record_relation.rbi
jeffcarbs: I left a more detailed comment but for this specifically:
- we're defining the methods directly on `ActiveRecord::Relation` so sorbet will use that
- `ActiveRecord::AssociationRelation` doesn't redefine `find` so when sorbet auto-generates the RBI is won't add that method to `ActiveRecord::AssociationRelation` and thus it will fallback to the inherited method and sig on `ActiveRecord::Relation`
- `ActiveRecord::Associations::CollectionProxy` actually redefines `find` (I think it just winds up calling super but might do something slightly different) which cases sorbet to generate that method when auto-generating the RBI. This would cause sorbet to use _that_ unsigged method definition instead of the inherited one so we need to explicitly define it ourselves here.

manhhung741:
> The sorbet_test_cases.rb now tests everything we're writing sigs for in those base classes (Base, Relation, etc) and that they're correctly typed on the model-specific classes so we basically just have to rely on that to know when we need to explicitly define methods vs falling back to the inherited ones.

Totally agree. It's best relying on exhaustive testing to verify the correctness of the generation logic. It's a lot of effort to reason about what should be generated, what would be overridden by sorbet.

manhhung741:
Follow up to #312 and #313. The first two commits in this PR are from the first one one, the next two are from the second.

This PR moves the `finder` methods from directly in the model RBIs to being generated in a base class (usually `ActiveRecord::Relation` although some needed to be explicitly defined in `CollectionProxy`).

Going by commit:
- Remove usage of `typed-override` in specs (pulled out into #315)
- Added missing assertions for the "numbered finders" (e.g. first, second, third)
- Remove the generation of most `finder_methods` from the model RBIs
- Regenerate the model RBIs

PR #316, opened 2020-04-26
Move more methods into base AssociationRelation and CollectionProxy RBIs

jeffcarbs:
~Follow up to #312, #313 and #314. The first two commits in this PR are from the first one one, the next two are from the second, next 4 are from the third.~ Rebased, so now this is a follow to just #314. The first three commits in this PR are from that one.

Move the last remaining methods from the `active_record_finder_methods` model plugin to be defined in the `Base` and `Relation` RBIs and remove the plugin.


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -69,16 +71,7 @@ def generate_active_record_relation_rbi
         return_type: "T::Array[T.type_parameter(:U)]",
       )
 
-      class_rbi.create_method(
-        "exists?",
-        parameters: [ Parameter.new("conditions", type: "T.untyped", default: "nil") ],
-        return_type: "T::Boolean",
-      )
-      class_rbi.create_method('any?', return_type: "T::Boolean")
       class_rbi.create_method('empty?', return_type: "T::Boolean")
```
jeffcarbs, line 74: `empty?` is actually only available on the relation classes, not on the model itself. All the others (exists?, any?, etc) are all available on both the model and the relation classes.

manhhung741: O.o 
Can you add your message as a comment in the code?

lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -151,13 +144,20 @@ def generate_active_record_relation_rbi
             return_type: "T.nilable(Elem)",
           )
         end
+
+        boolean_methods = %w(any? many?)
```
manhhung741, line 148: Wow, more special cases for 5.0. Why I'm not surprised.

PR #317, opened 2020-04-27
Remove some enumerable methods from model RBIs and rely on Base/Relation

jeffcarbs:
Remove some enumerable methods from model RBIs. In most cases these just fallback to the ones defined on `Base` and `Relation` although for some like `CollectionProxy` on certain Rails versions we need to manually define them in the RBI.


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -154,6 +169,10 @@ def generate_active_record_relation_rbi
         )
       end
 
+      if Rails.version =~ /^5\.(0|1)/
```
manhhung741, line 172: Fun time! lol

lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -105,6 +106,20 @@ def generate_active_record_relation_rbi
         value: "type_member(fixed: T.untyped)",
       )
 
+      # This _should_ work which would let us remove it from the enumerable_collections
+      # plugin but sorbet has a bug with T.any and generics.
+      # See: https://github.com/sorbet/sorbet/issues/2938
```
manhhung741, line 111: 😞

PR #321, opened 2020-04-29
Improve performance of model RBI generation

jeffcarbs:
This PR does two things that improve performance and memory usage:
- Write the model RBI after each loop. Previously we were generating all the model RBIs and storing the strings inside of an array and then once we generated it all we were writing it out to files. This is pretty inefficient from a memory perspective which likely leads to a performance hit as well.
- Speed up the `ActiveRecordNamedScope` plugin. I instrumented the plugins (just took `Time.now` before and after and printed it out) and was getting an average of ~4.5s for the ActiveRecordNamedScope. I found the culprit (`method_obj.owner.name` was very expensive for some reason 🤷) and took the time down to an average of 0.03s 🔥.

I ran this against my company's codebase with ~750 models.

**Before**
Using `0.6.3` I actually never got a full time because it was taking too long so I killed it. It was definitely longer than 10 minutes

**After**
Using this branch:
```
$ time bin/rails rails_rbi:all
... output ...

real	3m23.620s
user	2m37.853s
sys	0m21.239s
```

### DIFF
--- a/lib/sorbet-rails/model_plugins/active_record_named_scope.rb
+++ b/lib/sorbet-rails/model_plugins/active_record_named_scope.rb
@@ -7,15 +7,23 @@ class SorbetRails::ModelPlugins::ActiveRecordNamedScope < SorbetRails::ModelPlug
   def generate(root)
     model_class_rbi = root.create_class(self.model_class_name)
 
-    @model_class.methods.sort.each do |method_name|
+    # Named scope methods are dynamically defined by the `scope` method so their
+    # source_location is `lib/active_record/scoping/named.rb`. So we find scopes
+    # by two criteria:
+    # - they are defined in 'activerecord/lib/active_record/scoping/named.rb'
+    # - they are not one of the methods actually defined in that file's source.
+    #
+    # These methods are defined within `lib/active_record/scoping/named.rb` and should
+    # be ignored for the purposes of this plugin.
+    # See: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/scoping/named.rb
+    non_dynamic_methods = %i(all scope_for_association default_scoped default_extensions scope valid_scope_name?)
+
+    (@model_class.methods - non_dynamic_methods).sort.each do |method_name|
       next unless SorbetRails::Utils.valid_method_name?(method_name.to_s)
+
       method_obj = @model_class.method(method_name)
       next unless method_obj.present? && method_obj.source_location.present?
-      # we detect sscopes defined in a model by 2 criteria:
-      # - they don't have an owner name
-      # - they are defined in 'activerecord/lib/active_record/scoping/named.rb'
-      # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/scoping/named.rb
-      next unless method_obj.owner.name == nil
+
       source_file = method_obj.source_location[0]
       next unless source_file.include?("lib/active_record/scoping/named.rb")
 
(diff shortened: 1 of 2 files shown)


lib/sorbet-rails/model_plugins/active_record_named_scope.rb
```diff
@@ -7,15 +7,23 @@ class SorbetRails::ModelPlugins::ActiveRecordNamedScope < SorbetRails::ModelPlug
...
+    (@model_class.methods - non_dynamic_methods).sort.each do |method_name|
       next unless SorbetRails::Utils.valid_method_name?(method_name.to_s)
+
       method_obj = @model_class.method(method_name)
       next unless method_obj.present? && method_obj.source_location.present?
-      # we detect sscopes defined in a model by 2 criteria:
-      # - they don't have an owner name
-      # - they are defined in 'activerecord/lib/active_record/scoping/named.rb'
-      # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/scoping/named.rb
-      next unless method_obj.owner.name == nil
```
manhhung741, line 18: Who would have thought that finding the "owner" taking so much time...

lib/sorbet-rails/model_plugins/active_record_named_scope.rb
```diff
@@ -7,15 +7,23 @@ class SorbetRails::ModelPlugins::ActiveRecordNamedScope < SorbetRails::ModelPlug
...
+    # Named scope methods are dynamically defined by the `scope` method so their
+    # source_location is `lib/active_record/scoping/named.rb`. So we find scopes
+    # by two criteria:
+    # - they are defined in 'activerecord/lib/active_record/scoping/named.rb'
+    # - they are not one of the methods actually defined in that file's source.
+    #
+    # These methods are defined within `lib/active_record/scoping/named.rb` and should
+    # be ignored for the purposes of this plugin.
+    # See: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/scoping/named.rb
+    non_dynamic_methods = %i(all scope_for_association default_scoped default_extensions scope valid_scope_name?)
```
manhhung741, line 19: I'm thinking how to do this more systematically, and realize we can use "instance_methods" to find these methods, eg:
```
irb(main):023:0> pp ActiveRecord::Scoping::Named::ClassMethods.instance_methods.sort
[:all, :default_extensions, :default_scoped, :scope, :scope_for_association]
=> [:all, :default_extensions, :default_scoped, :scope, :scope_for_association]
irb(main):024:0> pp ActiveRecord::Scoping::Named::ClassMethods.private_instance_methods.sort
[:valid_scope_name?]
=> [:valid_scope_name?]
irb(main):025:0> pp ActiveRecord::Scoping::Named::ClassMethods.protected_instance_methods.sort
[]
=> []
```

I was also thinking we could just exclude all methods defined in ActiveRecord::Base, but someone may try to define a scope at the base and let it be inherited by all subclasses. However, we can also exclude all "Object" instance_methods from the considered list. Both `Object.instance_methods` and `Object.private_instance_methods` have a lot of methods.

lib/sorbet-rails/model_plugins/active_record_named_scope.rb
```diff
@@ -7,15 +7,23 @@ class SorbetRails::ModelPlugins::ActiveRecordNamedScope < SorbetRails::ModelPlug
...
+
+    (@model_class.methods - non_dynamic_methods).sort.each do |method_name|
```
manhhung741, line 21: Awesome. See my comment above about how to filter the method list

lib/sorbet-rails/model_plugins/active_record_named_scope.rb
jeffcarbs: Good call! Updated to dynamically get the files.

As for filtering out `Object.methods` or `Object.instance_methods`, it doesn't move the needle too much and could potentially cause issues if people define named scopes that collide so probably best to avoid for now.

PR #322, opened 2020-04-29
Loosen assoc_should_be_untyped check

jeffcarbs:
I had a model with associations like:
```
has_many :liabilities
has_many :payroll_taxes, through: :liabilities, source: :liability, source_type: 'PayrollTax'
```

This means that payroll_taxes is a "has many through" association where the join table is polymorphic so sorbet-rails was generating:
```ruby
sig { returns(ActiveRecord::Associations::CollectionProxy) }
def payroll_taxes; end
```

I would expect:
```ruby
sig { returns(::PayrollTax::ActiveRecord_Associations_CollectionProxy) }
def payroll_taxes; end
```

This was because it was using logic when checking if the reflection was polymorphic that said "is this a has-many-through? then check if the join table is polymorphic". I don't quite get that logic. In this case it shouldn't matter that the join table is polymorphic (in fact, I actually hadn't realized it was until this came up 😬).

Maybe there's a case you've run into that I'm not considering? If so we should figure out how to handle both ideally 🙂 If not then this PR simplifies the logic a bit.

### DIFF
--- a/lib/sorbet-rails/model_plugins/active_record_assoc.rb
+++ b/lib/sorbet-rails/model_plugins/active_record_assoc.rb
@@ -130,7 +130,15 @@ class SorbetRails::ModelPlugins::ActiveRecordAssoc < SorbetRails::ModelPlugins::
 
   sig { params(reflection: T.untyped).returns(T.nilable(T::Boolean)) }
   def assoc_should_be_untyped?(reflection)
-    polymorphic_assoc?(reflection) || !@available_classes.include?(reflection.klass.name)
+    # For some polymorphic associations (e.g. a has-many-through where the `source`
+    # is polymorphic) we can figure out the type from the class_name or source_type.
+    polymorpic_with_unknowable_klass = (
+      polymorphic_assoc?(reflection) &&
+      !reflection.options.key?(:class_name) &&
+      !reflection.options.key?(:source_type)
+    )
+
+    polymorpic_with_unknowable_klass || !@available_classes.include?(reflection.klass.name)
   end
 
   sig { params(reflection: T.untyped).returns(T.nilable(T::Boolean)) }


manhhung741:
So I think when it's polymorphic, we don't know the possible type and thus use "T.untyped. In this case, I think your code specifies a type (`source_type: "PayrollTax"`). I think that is a better signal about the type of the assoc. Perhaps the change should be using `source_type` etc when they are available.

lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -131,19 +131,12 @@ def populate_collection_assoc_getter_setter(assoc_module_rbi, assoc_name, reflec
...
   def relation_should_be_untyped?(reflection)
     # only type the relation we'll generate
     assoc_should_be_untyped?(reflection) || !@available_classes.include?(reflection.klass.name)
   end
-
-  sig { params(reflection: T.untyped).returns(T.nilable(T::Boolean)) }
-  def polymorphic_assoc?(reflection)
-    reflection.through_reflection ?
-      polymorphic_assoc?(reflection.source_reflection) :
-      reflection.polymorphic?
```
manhhung741, line 147: This recursive logic is to determine the polymorphic-ness of all assoc in the chain. I think it is still valid and we should keep it. I definitely remember running into a case that needed it and had to implement the recursive check.

jeffcarbs: Interesting.

When I was implementing this functionality in our model RBI rake task the issue that prompted me to handle the polymorphic case is that `reflection.klass` blows up if the `klass` can't be determined. I'm assuming that's why we're trying to be more thorough here to determine if this can be polymorphic.

I actually wound up taking a more empirical approach any just catching `reflection.klass` blowing up and setting the assoc_class to `T.untyped` in that case. I've updated the PR to use the approach more similar to the one we used in our rake task.

lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -19,22 +19,31 @@ def generate(root)
     model_class_rbi.create_include(assoc_module_name)
 
     @model_class.reflections.sort.each do |assoc_name, reflection|
+      begin
+        # TODO allow people to specify the possible values of polymorphic associations
+        assoc_class = reflection.polymorphic? ? "T.untyped" : "::#{reflection.klass.name}"
```
jeffcarbs, line 24: Playing around with this more, I think this is the most robust way to handle this problem. Our main goal is to always know the type if possible and ideally our solution for doing so would be resilient to changes within Rails (e.g. implementation changes, new options added, etc). Consider the following:
```ruby
has_many :pets, through: :pet_ownerships, class_name: 'Dog', source: :baz, source_type: 'Xyz'
```

Different permutations of those options will yield a different class that Rails will use for this association. In Rails 6 it looks for the class in the following order:
1. `class_name:` (https://github.com/rails/rails/blob/master/activerecord/lib/active_record/reflection.rb#L166)
2. `source_type:` (https://github.com/rails/rails/blob/master/activerecord/lib/active_record/reflection.rb#L1001-L1004)
3. `source_reflection.class_name` (same link)
    - I think this may have been the issue you ran into. If this were a reflection where `source` was polymorphic and it didn't specify either the `class_name` or `source_type` option then it would fallback to trying to call `source.klass.name` which would blow up because it couldn't know the type

After this digging, we _could_ apply that logic to the existing checks but it feels a little bad reaching into the implementation details of `reflection.klass` to know if it's safe to call it vs just calling it and rescuing.

Thoughts?

lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -128,22 +140,4 @@ def populate_collection_assoc_getter_setter(assoc_module_rbi, assoc_name, reflec
       return_type: nil,
     )
   end
-
-  sig { params(reflection: T.untyped).returns(T.nilable(T::Boolean)) }
-  def assoc_should_be_untyped?(reflection)
-    polymorphic_assoc?(reflection) || !@available_classes.include?(reflection.klass.name)
```
jeffcarbs, line 134: A test is now breaking (https://github.com/chanzuckerberg/sorbet-rails/blob/master/spec/model_rbi_formatter_spec.rb#L40-L55) that is testing certain classes not being in the `@ available_classes` set. Am I correct in assuming this test is meant to classes that no longer exist or can't be loaded for some reason?

If so, I think the test might not actually be tested what it's intending to. If the class were truly not in the codebase (e.g. `Message` is removed from the codebase but there's still a `has_many :messages` association lying around) `reflection.klass` this line would raise a `NameError`. This PR handles that case by rescuing the `NameError` but now the test isn't working because you can't "fake" a model not being available.

I updated the specs to remove that test and instead add an association to `Wizard` referencing a model that doesn't exist to test this out.

manhhung741: >  Am I correct in assuming this test is meant to classes that no longer exist or can't be loaded for some reason?
Yeah, this test is for special scenarios when the class somehow is not loaded. We err to the side of being correct (T.untyped is safe) than being wrong

lib/sorbet-rails/model_plugins/active_record_assoc.rb
manhhung741: I think there is something about the "through:" connection that we tried to handle with the recursive lookup that this doesn't handle. I'll take another pass later to understand what we tried to do there.

manhhung741: > After digging into this we could apply that logic to the existing checks but it feels a little bad reaching into the implementation details of reflection.klass to know if it's safe to call it vs just calling it and rescuing.

I understand the rationale now. It could be a better way of handling it than recursing through the relation chain (and maybe faster). At the end of the day, we would trust `reflection.klass` anyway. I'll investigate that in a follow up (if I can find the original issue).

PR #323, opened 2020-04-29
Move query methods to module

jeffcarbs:
One more PR that could help significantly reduce the side of the generated RBI.

There's a comment in the codebase that describes the query methods:
> a relation querying method will be available on
> - model (as a class method)
> - activerecord relation
> - asocciation collection proxy
> - association relation
>
> in case (1) and (2), it returns a Model::ActiveRecord_Relation
> in case (3) and (4), it returns a Model::ActiveRecord_AssociationRelation

Given that, this PR defines two modules per model:
```
module SomeModel::QueryMethodsReturningRelation
end

module SomeModel::QueryMethodsReturningAssociationRelation
end
```

- For case (1), they need to be class methods so we `extend QueryMethodsReturningRelation`.
- For case (2), they need to be instance methods so we `include QueryMethodsReturningRelation` 
- For case (3) and (4) they both need be instance methods so we `include QueryMethodsReturningAssociationRelation`

This allows us to remove half of the query methods since they only need to be defined in twice instead of four times.

### DIFF
--- a/lib/sorbet-rails/model_rbi_formatter.rb
+++ b/lib/sorbet-rails/model_rbi_formatter.rb
@@ -75,8 +75,6 @@ class SorbetRails::ModelRbiFormatter
       superclass: "ActiveRecord::Relation",
     )
     model_relation_rbi.create_include(self.model_query_methods_returning_relation_module_name)
-    model_relation_rbi.create_extend("T::Sig")
-    model_relation_rbi.create_extend("T::Generic")
     model_relation_rbi.create_constant(
       "Elem",
       value: "type_member(fixed: #{model_class_name})",
@@ -87,8 +85,6 @@ class SorbetRails::ModelRbiFormatter
       superclass: "ActiveRecord::AssociationRelation",
     )
     model_assoc_relation_rbi.create_include(self.model_query_methods_returning_assoc_relation_module_name)
-    model_assoc_relation_rbi.create_extend("T::Sig")
-    model_assoc_relation_rbi.create_extend("T::Generic")
     model_assoc_relation_rbi.create_constant(
       "Elem",
       value: "type_member(fixed: #{model_class_name})",
@@ -99,8 +95,6 @@ class SorbetRails::ModelRbiFormatter
       superclass: "ActiveRecord::Associations::CollectionProxy",
     )
     collection_proxy_rbi.create_include(self.model_query_methods_returning_assoc_relation_module_name)
-    collection_proxy_rbi.create_extend("T::Sig")
-    collection_proxy_rbi.create_extend("T::Generic")
     collection_proxy_rbi.create_constant(
       "Elem",
       value: "type_member(fixed: #{self.model_class_name})",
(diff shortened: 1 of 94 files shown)


spec/generators/rails-template.rb
```diff
@@ -133,7 +135,7 @@ def wood_type
     class Wizard < ApplicationRecord
       validates :name, length: { minimum: 5 }, presence: true
       # simulate conditional validation
-      validates :parent_email, presence: true, if: -> { false }
```
jeffcarbs, line 136: Sorbet was complaining about `-> { false }` and kept trying to downgrade this file so I changed it to be a symbol that references a method.

spec/generators/sorbet_test_cases.rb
```diff
@@ -183,13 +187,18 @@
 T.assert_type!(spell_books.one?, T::Boolean)
 # T.assert_type!(spell_books.update_all(name: 'Fantastic Beasts'), Integer) # Ignored until we add support
 # T.assert_type!(spell_books.delete_all, Integer) # Ignored until we add support
-# CollectionProxy query also typed correctly!
+# Query methods
+T.assert_type!(spell_books.all, SpellBook::ActiveRecord_AssociationRelation)
+T.assert_type!(spell_books.recent, SpellBook::ActiveRecord_AssociationRelation) # Named scope
+T.assert_type!(spell_books.unclassified, SpellBook::ActiveRecord_AssociationRelation) # Enum scope
+T.assert_type!(spell_books.unclassified.recent, SpellBook::ActiveRecord_AssociationRelation)
+T.assert_type!(spell_books.unclassified.recent.unscoped, SpellBook::ActiveRecord_Relation) # Turns back into relation
```
jeffcarbs, line 195: This was actually broken since we treated it as any other query method and had it returning the `SpellBook::ActiveRecord_AssociationRelation`.

manhhung741: Thanks for identifying this edge case!

lib/sorbet-rails/model_utils.rb
```diff
@@ -73,9 +83,10 @@ def exists_class_method?(method_name)
       root: Parlour::RbiGenerator::Namespace,
       method_name: String,
       parameters: T.nilable(T::Array[::Parlour::RbiGenerator::Parameter]),
+      standard_query_method: T::Boolean,
     ).void
   }
-  def add_relation_query_method(root, method_name, parameters: nil)
+  def add_relation_query_method(root, method_name, parameters: nil, standard_query_method: false)
```
manhhung741, line 89: I think it's not obvious when to pass `standard_query_method` and when to use the default value. Can you add a comment about it? It seems it should be true for "Rails" provided query methods, and examples of non-standard methods are named scopes. Do I understand it correctly?

lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -74,6 +74,7 @@ def generate_base_rbi(root)
       self.model_relation_class_name,
       superclass: "ActiveRecord::Relation",
     )
+    model_relation_rbi.create_include(self.model_query_methods_returning_relation_module_name)
     model_relation_rbi.create_extend("T::Sig")
```
manhhung741, line 78: While you're at it, can you remove these `extend T::Sig` lines? based on @paracycle we don't need them in RBI files

PR #324, opened 2020-04-30
Stop infering `validate :on` as unconditional

masylum:
Recent version of sorbet-rails added a neat optimization to make
attributes non nilable if they had an "unconditional validation".
After updating we had some parts of the code report wrongly unreachable code.
This is becase we have some validations using `:on` which are detected
– wrongly – as unconditional.

Didn't add any test since I saw that only the `:if` clause was actually
tested.

### DIFF
--- a/lib/sorbet-rails/model_plugins/base.rb
+++ b/lib/sorbet-rails/model_plugins/base.rb
@@ -35,7 +35,8 @@ module SorbetRails::ModelPlugins
       model_class.validators_on(attribute).any? do |validator|
         validator.is_a?(ActiveModel::Validations::PresenceValidator) &&
           !validator.options.key?(:if) &&
-          !validator.options.key?(:unless)
+          !validator.options.key?(:unless) &&
+          !validator.options.key?(:on)
       end
     end
   end


lib/sorbet-rails/model_plugins/base.rb
```diff
@@ -35,7 +35,8 @@ def attribute_has_unconditional_presence_validation?(attribute)
       model_class.validators_on(attribute).any? do |validator|
         validator.is_a?(ActiveModel::Validations::PresenceValidator) &&
           !validator.options.key?(:if) &&
-          !validator.options.key?(:unless)
+          !validator.options.key?(:unless) &&
+          !validator.options.key?(:on)
```
manhhung741, line 39: This is reasonable, given the doc: https://apidock.com/rails/ActiveModel/Validations/ClassMethods/validate
When any option is used, it's a conditional validates. Should we just check that "options" exist? I think it's safer to err  to the lesser strict option

manhhung741: cc @jeffcarbs since he added this code.

jeffcarbs: I'm not sure that documentation is correct. You can have a `message` option for instance:
```ruby
validates :field_name, presence: { message: 'Some custom message' }
```

masylum: I'm based on a real use case where this removed a `T.nilable` when it shouldn't. The documentation is this one: https://guides.rubyonrails.org/active_record_validations.html#on

> You can also use on: to define custom contexts. Custom contexts need to be triggered explicitly by passing the name of the context to valid?, invalid?, or save.

```
class Person < ApplicationRecord
  validates :email, uniqueness: true, on: :account_setup
  validates :age, numericality: true, on: :account_setup
end
 
person = Person.new(age: 'thirty-three')
person.valid? # => true
person.valid?(:account_setup) # => false
```

So basically, this should be treated as `T.nilable` (if the database says so), since the validation will be evaluated on runtime.

manhhung741:
Cool. What I get from the comments is that `on` is a conditional check, we
should use that. At the same time, there are other options that isn’t
conditional; we should still check the 3 possible options here explicitly.

PR #326, opened 2020-05-04
Type `in_batches`

jeffcarbs:


lib/sorbet-rails/active_record_rbi_formatter.rb
```diff
@@ -264,7 +264,6 @@ def create_elem_specific_query_methods(class_rbi, type:, class_method:)
         ],
         return_type: "T::Enumerator[#{inner_type}]",
         class_method: class_method,
-        override: true,
```
manhhung741, line 267: Curious why the test pass before this change and after. I'd think it should fail in one of the two case. Am I missing something? 🤔

jeffcarbs: It's weird. In my codebase, I don't generate hidden definitions so I get a sorbet error in the AR RBIs saying:
```
Method ActiveRecord::Base#find_each is marked override but does not override anything
```

I was under the impression that `override` would complain when:
- you're not overriding anything
   - This is true
- the method you're adding a sig for is not marked as `abstract`
   - This is not the case, see [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Amodule%20Something%0A%20%20sig%20%7B%20void%20%7D%0A%20%20def%20y%0A%20%20end%0Aend%0A%0Aclass%20A%0A%20%20include%20Something%0A%0A%20%20sig%20%7B%20override.void%20%7D%0A%20%20def%20x%0A%20%20end%0A%0A%20%20sig%20%7B%20override.void%20%7D%0A%20%20def%20y%0A%20%20end%0Aend%0A)

So the `override` here has no positive effect but could have the negative effect of raising a type error so I think we should remove it.

PR #327, opened 2020-05-09
Handle UUIDs and have the _ids method return the correct type

connorshea:
There are cases where you can have an ID that's not an integer type, for example when using UUIDs.

In my `schema.rb` I have an `events` table that's defined like this:

```ruby
create_table "events", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
  t.bigint "user_id", null: false
  t.bigint "eventable_id", null: false
  t.datetime "created_at", precision: 6, null: false
  t.datetime "updated_at", precision: 6, null: false
  t.integer "event_category", null: false
  t.jsonb "differences"
end
```

The primary key is a UUID rather than a bigint (it was originally going to be separate tables and I did UUIDs so I could merge it/split it out again without ID conflicts, but that's not important).

Whereas before, the return type of `id` was `T.untyped`, now it knows it should be a String because it understands what UUIDs should be:

```ruby
module Event::GeneratedAttributeMethods
  sig { returns(String) }
  def id; end
end
```

And in Users - which have events associated with them - it used to declare `event_ids` would return `T::Array[Integer]` but now correctly has it typed as `T::Array[String]`.
```ruby
module User::GeneratedAssociationMethods
  sig { returns(::Event::ActiveRecord_Associations_CollectionProxy) }
  def events; end

  sig { returns(T::Array[String]) }
  def event_ids; end
end
```

I also regenerated the type signatures with `srb rbi update` and regenerated the expected files. I think Attachments and Blobs' IDs methods are returning as untyped because the `schema.rb` doesn't actually include tables for ActiveStorage? The ActiveStorage methods have fine signatures for me in my Rails app.

Rails 6.0.3 changes type signatures for a lot of the methods in `active_record_relation.rbi` to be compatible with Ruby 2.7's method declaration changes, so there are errors in 6.0's expected `srb tc` output.

For example:

```
sorbet/rbi/gems/activerecord.rbi:5136: Method `ActiveRecord::AssociationRelation#create!` redefined with argument `attributes` as a non-splat argument https://srb.help/4010
    5136 |  def create!(attributes = nil, &block); end
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    sorbet/rails-rbi/active_record_relation.rbi:149: The corresponding argument `args` in the previous definition was a splat argument
     149 |  def create!(*args, &block); end
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
```


lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -114,9 +114,16 @@ def populate_collection_assoc_getter_setter(assoc_module_rbi, assoc_name, reflec
       return_type: relation_class,
     )
     unless assoc_should_be_untyped?(reflection)
+      if reflection.klass.table_exists?
+        # Normally the id_type is an Integer, but it could be a String if using
+        # UUIDs.
+        id_type = SorbetRails::ModelPlugins::ActiveRecordAttribute.new(reflection.klass, @available_classes).type_for_column_def(reflection.klass.columns_hash['id']).to_s
+      else
+        id_type = "T.untyped"
+      end
```
manhhung741, line 123: why don't we put this method into a base module shared between the plugins, eg `model_utils.rb`?

manhhung741:
@connorshea I'm thinking about cutting a release to fix Rails 6.0. Can you update this PR if you want it to be part of the release?

connorshea:
@manhhung741 I won't be able to finish this up until tomorrow night, so if you want to cut a release before then I'm fine with waiting until the next update.

manhhung741:
Ok. We’ll see.
I mainly want to get the next release out that removes fetch_typed haha. So
I want to create some buffer time between the two releases.

TobiasBales:
Any updates on this?

And btw: thank you for the awesome gem!

manhhung741:
@connorshea Hi Connor, you have time to give this PR an update?

manhhung741:
@TobiasBales Thanks :D I'm glad you find this gem useful! That gives me motivation to keep working on it.

manhhung741:
Connor already did all the base work. I'll take a stab at cleaning this PR up for merging.

manhhung741:
Tested in my repo that it generates the correct signature in a repo with postgres db

Issue #335, opened 2020-05-22
GeneratedUrlHelpers does not implement ActiveSupport::Concern so url helpers don't work


dtropp:
**Describe the bug:**
The provided GeneratedUrlHelpers module which wraps the Rails `Rails.application.routes.url_helpers` module stops the module binding to the class correctly.

Gives this error:
```
Started GET "/guidelines/4"
Processing by Guidelines::GuidelinesController#show as HTML
  Parameters: {"id"=>"4"}
NameError (undefined local variable or method `default_url_options' for #<Guidelines::GuidelinePresenter:0x00007f8ec04c7880>
Did you mean?  default_url_options=
               @@default_url_options)

NameError - undefined local variable or method `default_url_options' for #<Guidelines::GuidelinePresenter:0x00007f8ec04c7880>
Did you mean?  default_url_options=
               @@default_url_options:
  app/concepts/guidelines/guideline_presenter.rb:42:in `block in topics_raw'
  app/concepts/guidelines/guideline_presenter.rb:36:in `topics_raw'
  app/concepts/guidelines/guideline_presenter.rb:26:in `raw'
  app/concepts/guidelines/guideline_presenter.rb:16:in `to_json'
  app/concepts/guidelines/guidelines_controller.rb:11:in `show'
```
**Steps to reproduce:**
```
class GuidelinePresenter
    extend T::Sig
    include GeneratedUrlHelpers

   def my_url
     new_guideline_path()
   end
end
```

**Expected behavior:**
Generates the path using the rails url helper, without needing to implement `default_url_options()`.

Workaround:
```
# typed: false

  class GuidelinePresenter
    extend T::Sig
    UrlHelpers = Rails.application.routes.url_helpers
    include UrlHelpers

    def my_url
      new_guideline_path()
    end
  end
```

**Versions:**

- Ruby: ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin18]
- Rails: Rails 5.2.4.3
- Sorbet: Sorbet typechecker 0.5.5666 git f994f9d74e5ae379c76714a9b13872ece96ec3a1 debug_symbols=true clean=1
- Sorbet-Rails: sorbet-rails (0.6.5.1)

manhhung741:
Oh interesting. Let me double-check where `default_url_options` is coming from.

manhhung741:
Turns out this work for the "_path" method but not the "_url" method.
```
[...]

PR #338, opened 2020-05-23
Add default url options to url helpers

manhhung741:
Turns out classes including `url_helpers` does not need `default_url_options` defined explicitly to work (See #335). This PR implements a `default_url_options` method that draw it from the Rails application config so that people don't have to implement it explicitly.

Note that the default for `Rails.application.config` is {} and usually applications will set it for their mailer etc. The `_path` methods will work empty options {} but the `_url` methods still require a valid set of options with a HOST.

### DIFF
--- a/lib/sorbet-rails/rails_mixins/generated_url_helpers.rb
+++ b/lib/sorbet-rails/rails_mixins/generated_url_helpers.rb
@@ -11,6 +11,5 @@
 # + include GeneratedUrlHelpers
 # end
 #
-module GeneratedUrlHelpers
-  include Rails.application.routes.url_helpers
-end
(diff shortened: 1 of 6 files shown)


manhhung741:
@dtropp let me know how you think about this implementation :-)

dtropp:
> @dtropp let me know how you think about this implementation :-)

Yes, I think that's ok as a simple implementation. 

I was wondering if there was also an alternative using https://api.rubyonrails.org/classes/ActiveSupport/Concern.html, but I can't quite work it out.

manhhung741:
There is a simpler implementation that I just do this:
```
GeneratedUrlHelpers = Rails.application.routes.url_helpers
```
and it works.

dtropp:
Haha. Yes, I was doing that locally actually to get around Sorbet syntax rule error. Great!

PR #346, opened 2020-05-29
Add missing Kaminari method

ghiculescu:


manhhung741:
Ouch, looks like there are more errors due to hidden-definition.rbi. Maybe we should just remove that from sorbet folder. It isn't that useful for us and cause low signal build errors too often.

manhhung741:
I fixed the build error on master. Hopefully this PR will pass build check now 🤞

ghiculescu:
thank you!

PR #367, opened 2020-08-04
Upgrade to latest parlour

ghiculescu:
Bumping from 2.1 to 4.0.1: https://github.com/AaronC81/parlour/blob/master/CHANGELOG.md

I'm not sure if removing the rbi generated by `srb rbi gems` is the right approach (https://github.com/chanzuckerberg/sorbet-rails/pull/367/commits/86b820419c6b8491eb6137247c18bc41a5a59ffe), but I did it because of these `srb tc` fails: https://travis-ci.com/github/chanzuckerberg/sorbet-rails/jobs/368671021


ghiculescu:
Thank you

jsmestad:
@manhhung741 looks like the latest parlour causes issues with `srb init` you can see the issue here: AaronC81/parlour#110

Maybe it would make sense to loosen the dependency in the short term or yank the version?

manhhung741:
@jsmestad I'm torn because it looks like a problem with a new version of sorbet, not with the parlour gem. I feel like using a version of sorbet when it's not broken is a better fix (since sorbet is updated with a new version with each commit, you are not likely to lose much by using a recently released, but not the latest version)

PR #369, opened 2020-08-10
fix: primary key is not always "id"

dgodd:
Given the primary key for a table is not always "id", use reflection.class.primary_key instead.

I noticed this due to having two tables in my application with non-standard primary keys (Sadly they are very hard to change).

NOTE: I have not added a test for this since I'm not sure what would be most appropriate. I started changing "spell_books" to have a primary key of "spell_id" (in only rails v5.2 - and guessing you would want all versions changed). However, I figured it was better to ask for the preferences of the maintainers before continuing with more work.


smerchek:
Agree! My rails project also does not use `id` as the primary key.

manhhung741 (approved):
Sounds good! Thanks!

manhhung741:
Thank you @dgodd and sorry that it took awhile for me to get to this PR

PR #371, opened 2020-08-14
ActiveSupport -> to_{date,time,datetime} can return nil

giovannibonetti:
Hi, there!

This fixes some sigs for a couple of methods generated by ActiveSupport. This methods can return `nil` in some cases:
```ruby
''.to_date # nil
''.to_time # nil
''.to_datetime # nil
```


giovannibonetti:
Woops, I've opened the PR in the wrong repository. I'll open it on https://github.com/sorbet/sorbet-typed instead.

PR #375, opened 2020-09-16
Fix Plugin ActiveRecordAssoc undefined method oid for nil

donaldong:
### Steps to repro
Create an application record that is a view. Run `rake rails_rbi:models`. And see:
```
!!! Plugin SorbetRails::ModelPlugins::ActiveRecordAssoc threw an exception: undefined method `oid' for nil:NilClass
```

### Solution
For DB views, the id/pk column would not exist. So we should check for it first before using it to generate sigs. If the column does not exist, we don't want to generate any sigs.

### Test Plan
With this patch,  run `rake rails_rbi:models`; see no errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ rvm:
 matrix:
   include:
   - rvm: 2.5
-    env: RAILS_VERSION=5.2 SORBET_VERSION=0.5.5737
+    env: RAILS_VERSION=5.2 SORBET_VERSION=0.5.6210
   exclude:
   - rvm: 2.3
     env: RAILS_VERSION=6.0
--- a/spec/sorbet_spec.rb
+++ b/spec/sorbet_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 require 'open3'
 
-RSpec.xdescribe 'sorbet' do
+RSpec.describe 'sorbet' do
   let!(:harry) do
     Wizard.create!(
       name: 'Harry Potter',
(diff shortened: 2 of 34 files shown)


lib/sorbet-rails/model_plugins/active_record_assoc.rb
```diff
@@ -138,13 +138,18 @@ def populate_collection_assoc_getter_setter(assoc_module_rbi, assoc_name, reflec
...
     unless assoc_should_be_untyped?(reflection)
-      if reflection.klass.table_exists?
-        # Normally the id_type is an Integer, but it could be a String if using
-        # UUIDs.
-        id_type = type_for_column_def(reflection.klass.columns_hash[reflection.klass.primary_key]).to_s
-      else
-        id_type = "T.untyped"
-      end
+      id_column = reflection.klass.columns_hash[reflection.klass.primary_key]
+      return if id_column.nil?
```
manhhung741, line 142: @donaldong it shouldn't return here as that halts the method execution. It should only skip the generation of the "_ids" method.
```suggestion
      unless assoc_should_be_untyped?(reflection) && (id_column = reflection.klass.columns_hash[reflection.klass.primary_key])
         ...
      end
```

manhhung741 (commented):
@donaldong can you help me update this PR?

lib/sorbet-rails/model_plugins/active_record_assoc.rb
donaldong: Ah, thanks!

PR #378, opened 2020-09-21
ActiveJob#perform_now & #perform_later return instance of Job not void

xcskier56:
### Motivation
In ActiveJob, the `perform_now` and `perform_later` methods return an instance of the job that you can then interact with. In my case I'm taking the returned job and pulling the `DelayedJob` Job ID off of it and storing it. Currently in Sorbet-Rails, the `#perform_now` and `#perform_later` methods have their return type set to void. This PR fixes this.


spec/support/v5.0/sorbet_test_cases.rb
```diff
@@ -312,7 +312,11 @@
 T.assert_type!(HogwartsAcceptanceMailer.notify(wizard), ActionMailer::MessageDelivery)
 
 # Jobs
-AwardHousePointHourglasses.perform_later(student: wizard, point: 10)
+T.assert_type!(
+  AwardHousePointHourglasses.perform_later(student: wizard, point: 10),
+  AwardHousePointHourglasses
+)
```
manhhung741, line 318: Can you update the test case in the generator file instead of the generated file? Once you updated it in the generator, please run `rake update_spec:sorbet_test_cases`

https://github.com/chanzuckerberg/sorbet-rails/blob/master/spec/generators/sorbet_test_cases.rb

manhhung741 (changes requested):
Thanks for the change. It looks pretty good. 
I have one requested change: please update the test case in the source generator file instead of the `support/*` files (the latter are meant to be generated). See my inline comment for detail

xcskier56:
@manhhung741, thank you for the feedback. I have updated the generator and regenerated the test cases.

manhhung741 (approved):
THANK YOU!

Issue #16, opened 2019-06-20
Add support for custom attribute type (Rails 5)


manhhung741:
Rails 5 add a new API that let users define a custom type for attributes in a model. Currently Sorbet-rails read the attribute type from the database definition.

Learn more about attributes api: https://blog.bigbinary.com/2018/12/11/rails-5-attributes-api.html

PR #385, opened 2020-12-19
Add support for `serialized` attributes

alexdunae:
This PR adds support for `serialize` to ActiveRecord models.  

There are various "coders" (Hash, JSON) that have different type signatures which I've captured.  

I've got this running happily in production in our Rails app and I think I did everything right for the test suite, but please let me know if I've missed something.  I ran Ruby 2.6 locally and things were working.

```
./spec/bin/run_all_specs.sh
bundle exec srb tc
```
are both good.

An aside: whenever I'd regenerate the test Rails apps from the template (e.g. `bundle exec rake update_spec:v6_0`) I would get a ton of changes to the `typed` sigils in pretty much every Ruby file.  Is there another way that I'm missing?

### DIFF
--- a/lib/sorbet-rails/config.rb
+++ b/lib/sorbet-rails/config.rb
@@ -54,6 +54,7 @@ module SorbetRails
         :active_record_named_scope,
         :active_record_querying,
         :active_relation_where_not,
+        :active_record_serialization,
         :active_record_attribute,
         :active_record_assoc,
         :custom_finder_methods,
--- a/lib/sorbet-rails/config.rb
+++ b/lib/sorbet-rails/config.rb
@@ -54,7 +54,7 @@ module SorbetRails
         :active_record_named_scope,
         :active_record_querying,
         :active_relation_where_not,
-        :active_record_serialization,
+        :active_record_serialized_attribute,
         :active_record_attribute,
         :active_record_assoc,
         :custom_finder_methods,
--- a/spec/generators/sorbet_test_cases.rb
+++ b/spec/generators/sorbet_test_cases.rb
@@ -305,6 +305,17 @@ T.assert_type!(Wizard::House::Hufflepuff, Wizard::House)
 T.assert_type!(Wizard::House::Ravenclaw, Wizard::House)
 T.assert_type!(Wizard::House::Slytherin, Wizard::House)
 
+# Serialization
+T.assert_type!(
+  wizard.owl_results,
+  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[T.untyped, T.untyped], Integer, String)
+)
+T.assert_type!(
+  wizard.patronus_characteristics,
+  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[String, T.untyped], Integer, String)
+)
+T.assert_type!(wizard.pets, T::Array[T.untyped])
+
 # Mythical plugin
 T.assert_type!(Wand.mythicals, T::Array[Wand])
 
--- a/spec/generators/sorbet_test_cases.rb
+++ b/spec/generators/sorbet_test_cases.rb
@@ -308,13 +308,17 @@ T.assert_type!(Wizard::House::Slytherin, Wizard::House)
 # Serialization
 T.assert_type!(
   wizard.owl_results,
-  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[T.untyped, T.untyped], Integer, String)
+  T.nilable(T.any(T::Hash[T.untyped, T.untyped]))
+)
+T.assert_type!(
+  wizard.newt_subjects,
+  T.nilable(T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[T.untyped, T.untyped], Integer, String))
 )
 T.assert_type!(
   wizard.patronus_characteristics,
-  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[String, T.untyped], Integer, String)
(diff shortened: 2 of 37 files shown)


spec/generators/sorbet_test_cases.rb
```diff
@@ -305,6 +305,17 @@
 T.assert_type!(Wizard::House::Ravenclaw, Wizard::House)
 T.assert_type!(Wizard::House::Slytherin, Wizard::House)
 
+# Serialization
+T.assert_type!(
+  wizard.owl_results,
+  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[T.untyped, T.untyped], Integer, String)
```
manhhung741, line 311: It looks like `owl_results` is serialized as a Hash. I'd expect the type here to be `T::Hash[T.untyped, T.untyped]`. Why do we also have other possible type like Integer, String, etc. here?
```
  serialize :owl_results, Hash
```

manhhung741: In the generated RBI, it seems the type is only `T.nilable(T::Hash[...])`
```
  sig { returns(T.nilable(T::Hash[T.untyped, T.untyped])) }
  def owl_results; end
```

spec/generators/sorbet_test_cases.rb
```diff
@@ -305,6 +305,17 @@
...
+)
+T.assert_type!(
+  wizard.patronus_characteristics,
+  T.any(T::Array[T.untyped], T::Boolean, Float, T::Hash[String, T.untyped], Integer, String)
+)
+T.assert_type!(wizard.pets, T::Array[T.untyped])
```
manhhung741, line 317: In the generated type, pets can be a nilable array. Should we assert a nilable type here?
```
  sig { returns(T.nilable(T::Array[T.untyped])) }
  def pets
```

manhhung741 (commented):
Thank you for your contribution! It looks great! I had some question w.r.t the test case added.

> An aside: whenever I'd regenerate the test Rails apps from the template (e.g. bundle exec rake update_spec:v6_0) I would get a ton of changes to the typed sigils in pretty much every Ruby file. Is there another way that I'm missing?

I'll look into this issue!

spec/generators/sorbet_test_cases.rb
alexdunae: Yes, indeed.  Thanks. I've fixed up the `nil` handling throughout and corrected that incorrect `Hash` type.

Locally, I'd created a version where some of the serialised columns where nullable and some not.  I ran the tests and got a bunch of errors with `Wizard.create!` etc... since they now expected some non-null columns. 

Rather than update all the test factories to pass (and making this PR bigger) I instead just let all serialised columns by nullable.  If you'd prefer that another way lmk and I'm happy to make some changes.

spec/generators/sorbet_test_cases.rb
alexdunae: Fixed :)

alexdunae:
Thanks for the review.  I've made some fixes and updated with master.

spec/generators/sorbet_test_cases.rb
```diff
@@ -305,6 +305,21 @@
 T.assert_type!(Wizard::House::Ravenclaw, Wizard::House)
 T.assert_type!(Wizard::House::Slytherin, Wizard::House)
 
+# Serialization
+T.assert_type!(
+  wizard.owl_results,
+  T.nilable(T.any(T::Hash[T.untyped, T.untyped]))
```
manhhung741, line 311:
```suggestion
  T.nilable(T::Hash[T.untyped, T.untyped])
```

manhhung741 (approved):
Looks good. I think we can remove the `T.any` from T::Hash case.

spec/generators/sorbet_test_cases.rb
alexdunae: (I didn't know about Github's "suggestion" tools so added this in a separate commit - https://github.com/chanzuckerberg/sorbet-rails/pull/385/commits/ebbde2e95d54d7e82893594b4ae330561aa56dbd)

alexdunae:
One last question from me before you merge -- I put all the serialisation logic in its own plugin since it's a fairly self-contained concern. When I finished I realised that it shares a lot of logic with `ActiveRecordAttribute` (unsurprisingly), and there's a case to be made to merging "serialization" logic into `ActiveRecordAttribute` instead of being its own plugin.

So I just wanted to check... are you happy having it as its own plugin? Or would you prefer it be bolted on to `ActiveRecordAttribute`?

lib/sorbet-rails/config.rb
```diff
@@ -54,6 +54,7 @@ def initialize
         :active_record_named_scope,
         :active_record_querying,
         :active_relation_where_not,
+        :active_record_serialization,
```
manhhung741, line 57: let's rename this to be clearer: `active_record_serialized_attribute`

manhhung741 (commented):
@alexdunae Good question. I've been asking myself what's the proper way to do this as well. I like that the code is in a separate module; it's a nice way to break down the logic. I'm slightly concerned that people don't know attribute-related generation code is in two different modules. The "next" can be not intuitive, especially when plugins are configurable. 

I think renaming the plugin should be enough for now, and we can encourage creating more attribute-related plugins, like enum-related code can be its own plugin, and there can be another for [custom attribute type](https://github.com/chanzuckerberg/sorbet-rails/issues/16)

lib/sorbet-rails/config.rb
alexdunae: Nice.  Makes sense to me.  I've renamed the plugin.

PR #386, opened 2020-12-21
pluck_to_tstruct - Use tstruct prop's default value if type isn't nilable and the plucked value is nil

DanielGilchrist:
Consider this scenario
```ruby
class TUserStruct < T::Struct
  const :id, Integer
  const :name, String, default: "Unknown"
end

User.create!(name: nil)

# Below would throw a runtime type error because `nil` would be passed to `name` when the type is `String`
User.all.pluck_to_tstruct(TA[TUserStruct].new)
```

This PR makes it so that in this case `pluck_to_tstruct` will use the default value for the `name` prop ("Unknown") instead of throwing a type error

I decided to only allow this behaviour if the type isn't nilable because if the type is nilable then I assume we'd be explicitly saying we don't want to use the default in that case I guess? Keen for thoughts

### DIFF
--- a/spec/pluck_to_tstruct_spec.rb
+++ b/spec/pluck_to_tstruct_spec.rb
@@ -82,6 +82,18 @@ RSpec.describe SorbetRails::PluckToTStruct do
     end
   end
 
+  class WizardWithDefaultParentEmailT < T::Struct
+    const :name, String
+    const :house, String
+    const :parent_email, String, default: "hagrid@hogwarts.com"
+  end
+
+  class WizardWithDefaultNilableParentEmailT < T::Struct
+    const :name, String
+    const :house, String
+    const :parent_email, T.nilable(String), default: "hagrid@hogwarts.com"
+  end
+
   shared_examples 'pluck_to_tstruct' do |struct_type, expected_values|
     it 'plucks correctly from ActiveRecord model' do
       plucked = Wizard.pluck_to_tstruct(TA[struct_type].new)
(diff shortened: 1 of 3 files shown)


DanielGilchrist:
These test failures are really confusing have no idea what's causing this 😵 Any ideas?

```
1) SorbetRails::PluckToTStruct uses default value if prop type is not nilable and has default value it should behave like pluck_to_tstruct plucks correctly from ActiveRecord model
     Failure/Error: expect(plucked).to match_array(expected_values)

       expected collection contained:  [<WizardWithDefaultParentEmailT house="Gryffindor", name="Harry Potter", parent_email="hagrid@hogwart...efaultParentEmailT house="Gryffindor", name="Hermione Granger", parent_email="hagrid@hogwarts.com">]
       actual collection contained:    [<WizardWithDefaultParentEmailT house="Gryffindor", name="Harry Potter", parent_email="hagrid@hogwart...efaultParentEmailT house="Gryffindor", name="Hermione Granger", parent_email="hagrid@hogwarts.com">]
       the missing elements were:      [<WizardWithDefaultParentEmailT house="Gryffindor", name="Harry Potter", parent_email="hagrid@hogwart...efaultParentEmailT house="Gryffindor", name="Hermione Granger", parent_email="hagrid@hogwarts.com">]
       the extra elements were:        [<WizardWithDefaultParentEmailT house="Gryffindor", name="Harry Potter", parent_email="hagrid@hogwart...efaultParentEmailT house="Gryffindor", name="Hermione Granger", parent_email="hagrid@hogwarts.com">]
     Shared Example Group: "pluck_to_tstruct" called from ./spec/pluck_to_tstruct_spec.rb:166
     # ./spec/pluck_to_tstruct_spec.rb:100:in `block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:54:in `block (2 levels) in <top (required)>'
```
image

ghiculescu:
@DanielGilchrist your tests are failing because you haven't overridden the equality operators on your new structs. See https://github.com/DanielGilchrist/sorbet-rails/blob/pluck-to-tstruct-defaults/spec/pluck_to_tstruct_spec.rb#L46

spec/pluck_to_tstruct_spec.rb
```diff
@@ -74,7 +74,37 @@ class WizardWithWandT < T::Struct
...
+      self == other
+    end
+  end
+
+  class WizardWithDefaultParentEmailT < T::Struct
+    const :name, String
+    const :house, String
+    const :parent_email, String, default: "hagrid@hogwarts.com"
+
+    def ==(other)
```
ghiculescu, line 90: Not a blocker, but could you please see if there's a way to extract this common logic into a module, to make these classes easier to write? Ideally we'd do:

```ruby
class WizardWithDefaultParentEmailT < T::Struct
  include ComparableWizardStruct
  
  const :name, String
  # ... other consts ...
end
```

DanielGilchrist: Good call [`c193528` (#386)](https://github.com/chanzuckerberg/sorbet-rails/pull/386/commits/c1935284d347e9e68efeebc8685ec89a46328518)

ghiculescu (approved):
Looks great to me.

What do you think @manhhung741 ?

spec/pluck_to_tstruct_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'rails_helper'
 require 'sorbet-rails/model_rbi_formatter'
+require "tstruct_comparable"
```
ghiculescu, line 3:
```suggestion
require 'tstruct_comparable'
```

for consistency with the rest of the file

manhhung741:
This seems reasonable to me. Although I can see it being confusing to users because "nil" value behave differently depending on whether the type is nilable & there is a default value, I think it's also a natural behavior. 

I put it here for documentation:
- nil value, nilable prop, no default => nil
- nil value, nilable prop, has default => nil
- nil value, not nilable prop, no default => error
- nil value, not nilable prop, has default => default

PR #393, opened 2021-01-14
add a debug mode env that plugin failure will abort model generation

manhhung741:


lib/sorbet-rails/model_rbi_formatter.rb
```diff
@@ -115,11 +115,11 @@ def generate_base_rbi(root)
     params(
       plugins: T::Array[Parlour::Plugin],
       generator: Parlour::RbiGenerator,
-      allow_failure: T::Boolean,
     ).
     void
   }
-  def run_plugins(plugins, generator, allow_failure: true)
+  def run_plugins(plugins, generator)
+    allow_failure = ENV["SBR_DEBUG_MODE"] == "true"
```
coreyja, line 122: :wave: Unsolicited and post merge feedback, but I _think_ this might be backwards. I think you probably want to allow failures if the ENV VAR is NOT set to `true`

I think as is this will change the default (no env var) case from allowing failure to NOT allowing failure

I came across this while trying to debug some generation I had, slightly unfortunately in the version with debug mode I no longer get the error so I can't confirm my suspicion that this might be backwards!

PR #400, opened 2021-02-03
Fix ShrinePlugin's handling of remote_url plugin

reidab:
Passing T.untyped directly instead of as a string causes a runtime type check in Parlour to fail.


reidab:
It doesn't really seem reasonable that adding some quotes would drop the coverage by 46%. 😉 

Is there something else busted in the CI setup?

manhhung741 (approved):

PR #409, opened 2021-04-28
add `last_page?` method to kaminari plugin

stevenwchien:
This PR adds the [`last_page?`](https://github.com/kaminari/kaminari/blob/c5186f5d9b7f23299d115408e62047447fd3189d/kaminari-core/lib/kaminari/models/page_scope_methods.rb#L82) method to the `kaminari_plugin.rb` file. This method didn't seem to fit with the pattern of the other methods generated by `add_relation_query_method`, since it does not return a `Model::RelationType` (can't chain). Not sure if this is the best way to generate this method, but I added generators for `Model::ActiveRecord_Relation`, `Model::ActiveRecord_AssociationRelation`, and `Model::ActiveRecord_Associations_CollectionProxy`

### DIFF
--- a/lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
+++ b/lib/sorbet-rails/gem_plugins/kaminari_plugin.rb
@@ -43,5 +43,25 @@ class KaminariPlugin < SorbetRails::ModelPlugins::Base
       return_type: "Integer",
       class_method: true,
     )
+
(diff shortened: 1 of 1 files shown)


manhhung741 (approved):
Seems fine to me :-) The `add_relation_query_method` is just a helper to create common methods more easily; one doesn't have to use it if it doesn't fit the usecase.

manhhung741:
@stevenwchen: can you update this PR? it's out of date with the master branch & I don't have the permission to edit it.

PR #427, opened 2021-06-20
Add plugin for AASM gem.

connorshea:
This adds a few autogenerated methods that are created when a model has an AASM state machine.

Gem: https://github.com/aasm/aasm

I've tested this with my rails app vglist and am confident that it works correctly.


connorshea:
Looks like this is failing because Sorbet is using endless ranges, which aren't supported on Ruby 2.4 or 2.5: https://github.com/sorbet/sorbet/issues/4265 / https://github.com/sorbet/sorbet/pull/4281

We could either wait for the upstream fix (unclear if it'll be merged) or just drop Ruby 2.4 and 2.5 since they're EOL anyway.

connorshea:
After trying to investigate why the signature for `Wizard.all.in_batches` changed (or, perhaps, why Sorbet only just recently started complaining about it not matching the asserted type in the test cases file), I have genuinely no idea. My best guess is that there was a change in Sorbet at some point in the last few weeks which made Sorbet understand the types better, or something? 🤷‍♂️ 

The change from `T::Enumerable[Wizard::ActiveRecord_Relation]` to `ActiveRecord::Batches::BatchEnumerator` for the return type of the `in_batches` method was made last May. No idea why this test was passing up until recently 🤷‍♂️

connorshea:
> Looks like this is failing because Sorbet is using endless ranges, which aren't supported on Ruby 2.4 or 2.5: [sorbet/sorbet#4265](https://github.com/sorbet/sorbet/issues/4265) / [sorbet/sorbet#4281](https://github.com/sorbet/sorbet/pull/4281)
> 
> We could either wait for the upstream fix (unclear if it'll be merged) or just drop Ruby 2.4 and 2.5 since they're EOL anyway.
> 
> CI is also failing for Rails 6.0 / Ruby 2.7, not entirely sure why that is.

Looks like the latest Sorbet version restores compatibility with 2.4 and 2.5, so this should be fixed if we re-run travis.

manhhung741:
@connorshea Thank you! I've fixed the tests (`ActiveRecord::Batches::BatchEnumerator`) and dropped ruby 2.4 and 2.5 from the test matrix, as well as Rails 5.0 and 5.1.

connorshea:
Looks like CI passes now :)

PR #440, opened 2021-08-02
Update README to reflect `serialize` support

alexdunae:
Serialize support was added in https://github.com/chanzuckerberg/sorbet-rails/pull/385


manhhung741:
Thank you @alexdunae!

PR #441, opened 2021-08-03
Add money-rails plugin

alexdunae:


alexdunae:
This is extracted from my production Rails app where it's been running happily for about six months.  I looked but didn't see any tests for other plugins. Is there a plugin testing strategy currently?

alexdunae:
Hi @manhhung741 - this is extracted from our production Rails app where it's been running happily for about six months. I looked but didn't see any tests for other plugins. Is there a plugin testing strategy currently?

PR #474, opened 2021-11-23
Create modules for AASM auto generated status constants

AKhoo:
This is an enhancement to the AASM gem plugin.

The AASM gem [auto-generates constants](https://github.com/aasm/aasm#auto-generated-status-constants) for each AASM status. Currently, the plugin does not account for this, so Sorbet is unable to resolve these auto-generated constants. This PR adds that functionality.


ghiculescu:
Thanks!

PR #478, opened 2022-01-13
📝  Update `README.md` for dynamic attribute methods

danielvdao:
### Context

Looks like `will_save_change_to_<attribute>?` -> `will_save_change_to_attribute?(<attribute>)` [Rails docs](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-will_save_change_to_attribute-3F) was missing. Let me know if this is incorrect - happy to close if so!


ghiculescu:
Thanks!

PR #482, opened 2022-02-08
Add support for `where.missing` with new custom finder method `where_missing`

DanielGilchrist:
Adds typing support for [`where.missing`](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing) which was added in Rails 6.1

- Opted to add a new method `where_missing` to `lib/sorbet-rails/rails_mixins/custom_finder_methods.rb` for better type safety (`where(some_attribute: "some_value").missing(:some_association)` would throw as with args `where` returns an `ActiveRecord::Relation` object and not a `ActiveRecord::QueryMethods::WhereChain` object)
- `where_missing` and its type defs are only available if the Rails version is 6.1 or higher

We should probably also do something similar with `where.not` but it'd be a breaking change. Might tackle it in a future PR


lib/sorbet-rails/rails_mixins/custom_finder_methods.rb
```diff
@@ -26,5 +26,13 @@ def find_by_id(id)
     def find_by_id!(id)
       find_by!(id: id)
     end
+
+    # where.missing is only available in Rails 6.1 and above
+    unless Rails.version =~ /^5\./ || Rails.version =~ /^6\.0/
```
ghiculescu, line 31:
```suggestion
    if Rails.version >= "6.1"
```

DanielGilchrist: Last commit was completely incorrect fixed here 🤦‍♂️ [`5a9b17d` (#482)](https://github.com/chanzuckerberg/sorbet-rails/pull/482/commits/5a9b17d8114f929a374d7ee4beaaf39a874747a3)

ghiculescu:
@DanielGilchrist the remaining failing tests are also failing on master. Would you be interested in having a look at them? Happy to merge this either way.

ghiculescu:
Okay, the failing tests now appear to be yours.

ghiculescu (approved):

PR #484, opened 2022-02-10
fix workflows for CI and gem publishing

ashkulz:
Also made a change so that the CI always runs on Mondays, so that you get to know sooner when there's a breaking change due to dependency updates rather than a few months later when someone makes a PR and you don't know what happened.


ashkulz:
It gets generated automatically by srb rbi.

PR #497, opened 2022-03-10
Disable autoreleases on master

ghiculescu:
Reverts https://github.com/chanzuckerberg/sorbet-rails/pull/465

While this gem's maintenance status is unclear, I'm not sure this is a great idea.

Additionally, using `ENV['GITHUB_RUN_NUMBER']` in the gemspec made it not possible to use this gem using a github ref in your gemfile.

### DIFF
--- a/sorbet-rails.gemspec
+++ b/sorbet-rails.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name          = %q{sorbet-rails}
-  s.version       = "0.7.#{ENV['GITHUB_RUN_NUMBER'] || 100}".strip
+  s.version       = "0.7.32"
   s.date          = %q{2019-04-18}
   s.summary       = %q{Set of tools to make Sorbet work with Rails seamlessly.}
   s.authors       = ["Chan Zuckerberg Initiative"]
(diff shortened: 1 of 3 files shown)


sorbet-rails.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name          = %q{sorbet-rails}
-  s.version       = "0.7.#{ENV['GITHUB_RUN_NUMBER'] || 100}".strip
+  s.version       = "0.7.32"
```
ghiculescu, line 3: this is the current version on https://rubygems.org/gems/sorbet-rails

ashkulz:
@ghiculescu why is the maintenance status for this gem unclear? I don't see an issue talking about this. Would appreciate a more clearer update :slightly_smiling_face:

ghiculescu:
> @ghiculescu why is the maintenance status for this gem unclear? I don't see an issue talking about this. Would appreciate a more clearer update 🙂

What I should have said is that it's unclear to me. It doesn't seem like @manhhung741 has been around reviewing PRs for quite a while (which is fine, he doesn't have to).

ashkulz:
Thanks for the clarification! Does that mean you're the only one who can merge PRs?

ghiculescu:
As far as I know, yes

PR #505, opened 2022-04-23
Allow manual gem publish via github actions

ghiculescu:


ashkulz:
@ghiculescu I'm suspecting that
```yaml
    - uses: actions/checkout@v2
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: 2.7
```
is required, as the action uses [`gem build`](https://github.com/dawidd6/action-publish-gem/blob/master/main.rb#L30) and it possibly can't find the correct Ruby? I'm not sure, that's my best guess on why the earlier workflow worked and the current doesn't :man_shrugging:

ashkulz:
@ghiculescu thinking it through I'm pretty sure of that, because without a checkout it won't find any gemspec files :grin: Not sure if the specific Ruby version is needed, as the default image should have ruby installed anyway :man_shrugging:

ghiculescu:
Haha yep that’ll do it. Thanks! That should be ready to go now.

PR #506, opened 2022-04-23
fix manual gem publishing workflow

ashkulz:


ashkulz:
@ghiculescu can you check if this works? :crossed_fingers:

PR #563, opened 2023-06-08
Zip plucked values with tstruct keys in the same order for `pluck_to_tstruct`

### LINKED ISSUE #562: `pluck_to_tstruct` assigns plucked values to the wrong tstruct keys depending on the order in which the tstruct props for associations are listed
**Describe the bug:**

When calling `pluck_to_tstruct` with association mappings, the plucked values do not get assigned to the correct tstruct keys if any of the tstruct props for the associations are listed before the non-association ones.

This happens because the associations always end up at the end of the `pluck_keys` array, due to the fact that we have to replace the association keys with their corresponding association values from the association mappings param:
https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/lib/sorbet-rails/rails_mixins/pluck_to_tstruct.rb#L33

So when the row of plucked keys gets zipped with `tstruct_keys`, the plucked values do not necessarily get matched up with their corresponding keys, since the association keys in `tstruct_keys` are in their initial positions rather than at the end of the `tstruct_keys` array.
https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/lib/sorbet-rails/rails_mixins/pluck_to_tstruct.rb#L40

In addition, this means that if the order of the association keys within the `associations` param to `pluck_to_tstruct` differs from the order of their corresponding keys in the tstruct class, the plucked association values will also get assigned to the incorrect tstruct keys.

**Steps to reproduce:**

Call `pluck_to_tstruct` with a tstruct class that lists the prop for an association mapping (`wand_wood_type`) before any of the non-association ones:

```
class WizardWithWandT < T::Struct
    include TStructComparable

    const :name, String
    const :wand_wood_type, String
    const :house, String
end

Wizard.joins(:wand).pluck_to_tstruct(TA[WizardWithWandT].new, associations: { wand_wood_type: "wands.wood_type" })
```

The plucked value for `wands.wood_type` gets assigned to `house` instead of `wand_wood_type` (which also means the plucked `house` gets assigned to `wand_wood_type`)

```
[
     <WizardWithWandT house="Holly" name="Harry Potter" wand_wood_type="Gryffindor">,
     <WizardWithWandT house="Vine" name="Hermione Granger" wand_wood_type="Gryffindor">,
]
```

OR

Call `pluck_to_tstruct` with a tstruct class that lists the props for association mappings in a different order than the keys in the `associations` param:

```
class WizardWithWandT < T::Struct
    include TStructComparable

    const :name, String
    const :house, String
    const :wand_wood_type, String
    const :wand_core_type, Integer
end

Wizard.joins(:wand).pluck_to_tstruct(
     TA[WizardWithWandT].new,
     associations: { wand_core_type: "wands.core_type", wand_wood_type: "wands.wood_type" },
)
```

It incorrectly tries to assign the plucked value for `wands.core_type` to `wand_wood_type`, and results in a TypeError:

```
Failure/Error: tstruct.new(value)

TypeError:
     Parameter 'wand_wood_type': Can't set WizardWithWandT.wand_wood_type to 0 (instance of Integer) - need a String
```

**Expected behavior:**

All the plucked values get correctly assigned to their corresponding tstruct keys.

```
[
     <WizardWithWandT house="Gryffindor" name="Harry Potter" wand_wood_type="Holly">,
     <WizardWithWandT house="Gryffindor" name="Hermione Granger" wand_wood_type="Vine">,
]
```

```
[
     <WizardWithWandT house="Gryffindor" name="Harry Potter" wand_core_type=0 wand_wood_type="Holly">, 
     <WizardWithWandT house="Gryffindor" name="Hermione Granger" wand_core_type=0 wand_wood_type="Vine">,
]
```

**Versions:**

- Ruby: 3.1.2
- Rails: 6.1.7.2
- Sorbet: 0.5.10860
- Sorbet-Rails: 0.7.34

gnaratil2017:
Fixes #562

### DIFF
--- a/lib/sorbet-rails/rails_mixins/pluck_to_tstruct.rb
+++ b/lib/sorbet-rails/rails_mixins/pluck_to_tstruct.rb
@@ -31,13 +31,14 @@ module SorbetRails::PluckToTStruct
     end
 
     pluck_keys = (tstruct_keys - associations_keys) + associations.values
+    tstruct_keys_in_pluck_order = tstruct_keys - associations_keys + associations_keys
 
     # loosely based on pluck_to_hash gem
     # https://github.com/girishso/pluck_to_hash/blob/master/lib/pluck_to_hash.rb
     keys_one = pluck_keys.size == 1
     pluck(*pluck_keys).map do |row|
       row = [row] if keys_one
-      value = Hash[map_nil_values_to_default(tstruct_props, tstruct_keys.zip(row))]
+      value = Hash[map_nil_values_to_default(tstruct_props, tstruct_keys_in_pluck_order.zip(row))]
       tstruct.new(value)
     end
   end
--- a/spec/pluck_to_tstruct_spec.rb
+++ b/spec/pluck_to_tstruct_spec.rb
@@ -58,8 +58,8 @@ RSpec.describe SorbetRails::PluckToTStruct do
     include TStructComparable
 
     const :name, String
-    const :house, String
     const :wand_wood_type, String
+    const :house, String
   end
 
   class WizardWithDefaultParentEmailT < T::Struct


spec/pluck_to_tstruct_spec.rb
```diff
@@ -58,8 +58,8 @@ class WizardWithWandT < T::Struct
     include TStructComparable
 
     const :name, String
-    const :house, String
     const :wand_wood_type, String
+    const :house, String
```
gnaratil2017, line 62: With this change in the order of the tstruct props, the spec on line 138 fails without the new changes to `pluck_to_tstruct.rb`.

https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/spec/pluck_to_tstruct_spec.rb#L138-L147

Rather than the expected array, the actual array would look like:

```
[
     WizardWithWandT.new(name: "Harry Potter", house: "Holly", wand_wood_type: "Gryffindor"),
     WizardWithWandT.new(name: "Hermione Granger", house: "Vine", wand_wood_type: "Gryffindor"),
]
```

lib/sorbet-rails/rails_mixins/pluck_to_tstruct.rb
```diff
@@ -31,13 +31,14 @@ def pluck_to_tstruct(ta_struct, associations: {})
...
     pluck_keys = (tstruct_keys - associations_keys) + associations.values
+    tstruct_keys_in_pluck_order = tstruct_keys - associations_keys + associations_keys
 
     # loosely based on pluck_to_hash gem
     # https://github.com/girishso/pluck_to_hash/blob/master/lib/pluck_to_hash.rb
     keys_one = pluck_keys.size == 1
     pluck(*pluck_keys).map do |row|
       row = [row] if keys_one
-      value = Hash[map_nil_values_to_default(tstruct_props, tstruct_keys.zip(row))]
+      value = Hash[map_nil_values_to_default(tstruct_props, tstruct_keys_in_pluck_order.zip(row))]
```
gnaratil2017, line 41: When we zip the tstruct keys with the plucked values, we want to make sure the corresponding key and value match up correctly.

This requires that our tstruct keys are in the same order as the `pluck_keys`, which is why we want to subtract the `associations_keys` and then append them to the end of `tstruct_keys` on line 34.

spec/pluck_to_tstruct_spec.rb
gnaratil2017, line 62: With this change in the order of the tstruct props, the spec on line 138 fails without the new changes to `pluck_to_tstruct.rb`.

https://github.com/chanzuckerberg/sorbet-rails/blob/447ecbe4a2586ce8aa9055290d571fe34a55e27f/spec/pluck_to_tstruct_spec.rb#L138-L147

Rather than the expected array, the actual array would look like:

```
[
     WizardWithWandT.new(name: "Harry Potter", house: "Holly", wand_wood_type: "Gryffindor"),
     WizardWithWandT.new(name: "Hermione Granger", house: "Vine", wand_wood_type: "Gryffindor"),
]
```

